PR #3, opened 2015-09-10
Changed standard settings

ReinerBforartists:
This branch brings the Bforartists standard settings. See tracker entry: http://www.bforartists.de/content/change-standard-settings


ReinerBforartists:
closed

Issue #843, opened 2019-05-02
Video Sequencer - Strip menu - Swap enumerated operators menu move up one


Draise14:
![bforartists_2019-05-02_11-55-55](https://user-images.githubusercontent.com/35357409/57092549-400bf180-6cd1-11e9-8d84-e4fa65ce1ae0.png)

## Task
Move the Sequencer Swap operators up one, so it's easier to access. 

## Issue
They are enumerated menu items, though you can make them into `layout.operator` items - though being enumerated. 
![atom_2019-05-02_11-53-50](https://user-images.githubusercontent.com/35357409/57092552-413d1e80-6cd1-11e9-8d75-0ef5dcf92cea.png)

ReinerBforartists:
Do it! :)

Draise14:
I have brought these up one level, renamed the tooltip, added icons (both in the Enumerated menu and in the python menu, and it's in the commit will be in the #765 task.

ReinerBforartists:
Please commit every single issue as a single commit. One big commit that contains everything is a nightmare to maintain and to follow.

ReinerBforartists:
Ah, and please make a separated pull request for every finished task. Not one pull request for everything. This puts me under pressure since the amount of code to check is much bigger then.

Work in small chunks!

Draise14:
Ah ok. How do I do this when the tasks are all in the same file? I probably should have finished the other task before even starting this one. I jumped the gun when I got assigned, as it was easy, lol

ReinerBforartists:
Yeah, now we have a problem it seems.

Not a happy bunny with that. As told, work in smallest possible chunks. But now it is as it is. We will see how we can handle the pull request then.

Draise14:
There was a merge regression here, so re-committing.

Draise14:
Ok, done. 

![bforartists_2019-05-21_12-38-31](https://user-images.githubusercontent.com/35357409/58118009-9039ed80-7bc5-11e9-9f6a-0237e34a2ab4.png)

Draise14:
I just realized when I do a pull request, all my commits on different tasks go to that one pull request. 

Does this mean that if a pull request is per task/commit ideally.. I have to pause work on any new tasks till the pull is resolved?

ReinerBforartists:
Yes. This would make my life muuuch easier when i just need to check one issue ...

Draise14:
[...]

Issue #765, opened 2019-04-06
Video Sequencer - Strip menu needs icons


ReinerBforartists:
We need some icons here.

![sequencerstripmenu](https://user-images.githubusercontent.com/13919702/55671909-2f648900-5895-11e9-9622-5cbbf37b225b.jpg)

Draise14:
Maybe I can learn how to put in icons with this task? I use the VSE a lot.

ReinerBforartists:
Have a look here :)

https://www.bforartists.de/data/tuts/Icontutorial.pdf

Draise14:
Hey cool. I added an Icon to the UI. It already existed so.. piece of cake! The tutorial is really helpful, bit all over the place - yet I like how more info is better. I wish everything was like this. 

![bforartists_2019-04-11_11-58-57](https://user-images.githubusercontent.com/35357409/55976250-56cea380-5c51-11e9-9a42-c80ff2058c16.png)

Looking at the sheet I've found a lot of design inconsistencies everywhere... but what a huge artistic feat it would be to design consistent icons across the board! 

Now I'll.. try put in existing icons first then create any new ones later.

ReinerBforartists:
Gratulations :)

Yes, There's a ton of inconsistency. Starts with the icon names already as you may found out with the copy icon already ^^

I think somebody is already working at a new icon sheet. But not sure. He is so quiet at the moment ^^

Draise14:
Oh there is someone working on it? The new design for all should be fun. 

Re-opened. Hang on, I haven't finished nor committed this one yet.

Draise14:
Who is assigned to the Icon redesign?

ReinerBforartists:
I hate the close button XD Sorry ^^

Somebody at the forum. Rayman. He hasn't created a github account yet. So not sure if he is still working at it. It's a monster task for sure.

Draise14:
Ah right, yeah, would make sense. I'll use another instance of Inkscape to do the creation, then copy/paste in. No issue. 

Would be great to see Rayman here on Github.

ReinerBforartists:
You can drag svg files directly into the sheet. No need for copy n paste.

ReinerBforartists:
Sorry for poking, but any progress here? The task is blocking all the other tasks where icons are missing Makes no sense to add new icons when you commit an old icon sheet then :)

Draise14:
[...]

Issue #846, opened 2019-05-03
Properties Editor - Preview - missing icons wrong alignment


ReinerBforartists:
Something is broken with the preview. Buttons at the right lacks of icons and are not aligned at the left as they should.

![previewprop problem](https://user-images.githubusercontent.com/13919702/57128483-d56fba00-6d93-11e9-9319-297e6b49fae9.jpg)

ReinerBforartists:
Let's have a look what is going on here ...

ReinerBforartists:
Okay, two tasks in one. The icons are missing. New introduced ones. Let's find out if we had old icon versions here.

ReinerBforartists:
Nice, adding the icons aligns the buttons at the right then it seems. Okay, let's grab the three new icons from Blender ...

ReinerBforartists:
Fixed in master

![matpreviewfixed](https://user-images.githubusercontent.com/13919702/57133554-71ed8880-6da3-11e9-9f61-71c23acd2a72.jpg)

PR #848, opened 2019-05-03
(2 tasks)  #765 and #843

Draise14:
Video Sequencer - Strip menu needs icons #765
Video Sequencer - Strip menu - Swap enumerated operators menu move up one #843


Draise14:
The image files have the new icons from #846, but the UI_icons.h file doesn't have them listed, so the DAT files for those icons have not been generated. 

inkscape_2019-05-03_11-30-41

ReinerBforartists:
So what is the actual change here? Is this this monster task?

Keep your hands away from the icons outside of the sheet. Those are icons that are not longer needed. But could still be good place holders. 

We will clean them up once we do the new icon sheet. For now simply ignore them.

Draise14:
This isn't that big, just icons for the tasks I did. #846 looks like it isn't merged in yet... I'll do that now.

ReinerBforartists:
NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

One commit one pull request we said :)

Draise14:
Ok... I'll let you do the merge later (I'll make a task). Just to say.. it might have been best to have paused any icon work till I committed and closed the icon work I was doing. 

I'll drop these for now, I have done my work on these tasks, the #846 icon work is still not part of this branch.

ReinerBforartists:
As told in the forum, i will not merge this before the new version is out. I'll do this without pressure. So please be patient a bit.

ReinerBforartists:
Can't accept this pull request, sorry. It breaks everything.

The iconsheet is killed. No icons in the svg file anymore. But odd descriptions that should not be in the sheet. The size fits, but my inkscape version shows nothing. Just curious, what inkscape version are you using? I have still version 0.9.2 here. Either you have an older version than that, or a newer one.

And it seems that the creation itself went also partially wrong. The png files misses the new icons, which should be up right as i can see from your shot and the code. There is none. Which means you did something wrong at one point. Those png files are intermediate steps. Which must contain the new icons.

And last but not least, with your code i can't access the video layout anymore. It closes with a visual basic runtime error. So you broke the code somewhere too :/

noiconsanymore

wherearetheicons

debugproblem

Draise14:
I used Inkscape 0.92.4, the latest stable version on their website. I'll redo this to get it right.

ReinerBforartists:
0.9.2.4 here too. So it's not the inkscape version.

Start with one small task if possible. We first need to nail down what goes wrong here.

ReinerBforartists:
The icon sheet is still empty, all icons gone ...

Draise14:
I merged the master and did a "conflict resolve" to make it revert to the master. The work is being repeated right now. I'll commit again with it compiling correctly later today.

ReinerBforartists:
Looking good now :)

One little thing, please don't use overlong names. This one is a bit too long for my flavour: SEQUENCER_RELOAD STRIPS AND ADJUST LENGTH . Yes i know, that's the tool name. But i would've most probably shortened it to SEQ_RELOAD_AND_ADJUST

We could run into a internal number of characters limit when this icon needs to be used in the C code.

Another thing is that i noticed that you use "inkscape label" . Nothing wrong with that. That's what the inkscape label is meant for. 

I personally simply change the id to the name. Makes the entry shorter, and the list better to read in the xml editor. And you will quickly notice when you have a double in the sheet. Since there can just be one unique id ... . A same label though can exist multiple times.

And you use the name in both in two cases, id and label. One could be cleaned up :)

Sorry for being pedantic here. It most probably works also this way. But would you be so kind to correct the above points. Now is the time to correct such little things. Before it's in the master :)

ReinerBforartists:
We have still a problem with a not accessible video editor layout ...

idandlabel

ReinerBforartists:
I am lost here with this runtime problem. No obvious cause to find. The code is nearly identical. It could have to do with the too long icon names. There could be a limit that we run into here.

I would say shorten the icon names down to let's say maximum 16 characters. Then let's have a second look at the runtime error.

Draise14:
Ok, reproduced the error. Even found the file that has the issue ("stripe_offs" is not initialized). I did only find this in **Debug** building mode though. 

A Release compilation doesn't give me the error. 

bforartists_2019-05-08_09-22-40

ReinerBforartists:
Is the error gone now after shorten the icon names? As told, i am as clueless as you. The code looks correct. The only real difference that counts is the icon names.

When it still exists, then you might need to go from scratch, with just one icon. And see at which point the error reappears.

ReinerBforartists:
Thanks for the hint with the debug build Draise. This error appears in the master too. Funny little rascal.

Okay, i will check your branch again, and merge it when everything looks okay. I will try to catch this error with a debug build error run. It's definitely another issue.

Hmm, how to make a debug build and run debug then would be of interest for you too. The mean part is, it's not necessarily the file in the error report that is the real trouble maker ...

ReinerBforartists:
Created a task for the crash

https://github.com/Bforartists/Bforartists/issues/864

ReinerBforartists:
Found one last thing. You should remove the lines 515 and 516 in the space_sequencer.py file. These lines are obsolete. And there is no reason to keep this code in the python part. When we keep every change here then we end in a mess. In case you want to follow a change then use the git history.

Different story in the C part. Here comments are vital so that the merges doesn't kill old changes. I don't have all vital changes in the head. But icon changes and changes at the tooltips should also here be save even without any comment.

As a side comment, i would put this enum into the top level anyways. Three hierarchies deep is one hierarchy too deep for my flavour. And there is enough space.

And i have to repeat my wish with the icon labeling in the blender_icons.svg. Please use the name in the id part, and remove the inkscape.label from your icons.

You don't need to recreate the icons then. The names are just relevant in the inkscape file, for easier navigation. I don't want to have our sheet as messy as the Blender iconsheet.

remove

delete

Draise14:
> As a side comment, i would put this enum into the top level anyways. Three hierarchies deep is one hierarchy too deep for my flavour. And there is enough space.

Task #843 is to do exactly that. I can commit those changes, sure. 

> And i have to repeat my wish with the icon labeling in the blender_icons.svg. Please use the name in the id part, and remove the inkscape.label from your icons.

>You don't need to recreate the icons then. The names are just relevant in the inkscape file, for easier navigation. I don't want to have our sheet as messy as the Blender iconsheet.

Wait.. but all of them have labels also, and even misnamed ones. 
The ID's currenlty have all the correct names the names, and the Label is the same, as generally the practice I noticed in the file has label as the name (but with a hashtag). 

inkscape_2019-05-08_11-50-58
inkscape_2019-05-08_11-51-08

Is it by putting in the Hashtag that it doesn't show up in the XML? Due to never touching XML in SVG files, I assumed the layer/object in the manager would show it's label (they don't show the ID automatically).

Draise14:
Ok, fixed it up and recommitted the cleaned up iconsheet, no labels now, AND I moved the enumerated swap strips up a level and cleaned that up.

ReinerBforartists:
Yes i know, there is old crap in there. Most of them old Blender icons. But even i made some mistakes in the past. You never stop learning :)

ReinerBforartists:
Done. Files are in master now. Many thanks for your patience :)

Issue #894, opened 2019-05-19
Video Sequencer - re-organize the Change Effect Type menu's to be consistent with Add menus


Draise14:
## Issue
![PureRef_2019-05-18_20-56-44](https://user-images.githubusercontent.com/35357409/57976835-935b9080-79af-11e9-9db7-1049657ab0c4.png)
When you compare the Add menu's for effect strips in the VSE, you get a very different order and system when you want to "Change Effect Type", not only are transition strips grouped with the effect strips, the order is all out of whack. There are even some entries that aren't there! 

## Solution
Re-ordering, splitting up, or adding dividers to make it consistent would be great - so they are all the one and the same order for muscle memory and consistency. 

## Alternatives
Another solution could be to divide the Change Effect Type menu to remove the Transition strips to have it's own menu, similar to the add menus.

ReinerBforartists:
Please wait with this changes a bit longer. There are changes at the VSE code incoming ...

Draise14:
Ah right, sure. What are they adding to these? 
Just curious, how do you add line breaks to enumerated menu's? I've had a look (one is built in Python, the other an enumerated C menu).

ReinerBforartists:
Line break is \n

ReinerBforartists:
To quote the dev meeting notes from tomorrow: 

> Sequencer menus and sidebar panels have been reorganized for consistency and clarity. (Peter Fog, William Reynish)

ReinerBforartists:
Flight we have green. The merge is in the master now ^^

Draise14:
Oh great.

Draise14:
![PureRef_2019-05-21_09-40-37](https://user-images.githubusercontent.com/35357409/58105700-8a83de00-7bac-11e9-9047-c5a001546945.png)

Ok.. the effect strip organization is now the same as the Change Effect Type, great. 

But now I see a huge inconsistency.. as you can change some effect strips into things that aren't even considered effects, eg. adjustment layer, text, color and transition strips. 

Trying to change these effects to these other ones creates errors if you do try change an effect strip, like they should be in their own enumerated menu to swap out their own kind (transition strips), or removed entirely. 

@ReinerBforartists , what would you do? 
Should I commit and close like this and make a new task with the issue of creating a new "Change Transitions" enumerated menu? Then add another task to remove the out of place "effects" content strips (are we sure we want to change an effect strip to a content strip as a feature?) 
[...]

PR #902, opened 2019-05-21
 Video Sequencer - re-organize the Change Effect Type menu's to be consistent with Add menus #894 

Draise14:
Video Sequencer - re-organize the Change Effect Type menu's to be consistent with Add menus #894 

- Added separators to enumerated menu to be a little more consistent without changing features
- Added icons to enumerated menu
- fixed icons to match in enumerated menu for consistency

### DIFF
--- a/release/scripts/startup/bl_ui/space_sequencer.py
+++ b/release/scripts/startup/bl_ui/space_sequencer.py
@@ -455,7 +455,7 @@ class SEQUENCER_MT_add(Menu):
 
         layout.separator()
 
-        layout.operator("sequencer.effect_strip_add", text="Adjustment Layer", icon='COLOR').type = 'ADJUSTMENT'
+        layout.operator("sequencer.effect_strip_add", text="Adjustment Layer", icon='NODE_DISPLACE').type = 'ADJUSTMENT'
 
         layout.operator_context = 'INVOKE_DEFAULT'
         layout.menu("SEQUENCER_MT_add_effect")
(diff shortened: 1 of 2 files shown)


Draise14:
Weird, task #843 that I re-opened and recommitted after a merge regression got lumped into this pull request automatically.

ReinerBforartists:
That's normal. There must be a way to commit changes to the pull request ...

Ideally please just make one pull request for one task. But sometimes it's not to avoid to have more than one task in a pull request. The problem is, it makes me more work to check then ...

I promise to merge it as fast as possible, so that you can continue :)

ReinerBforartists:
The second reason why just one commit for one issue is, when you need something to correct then you end in a big mess.

Found a small thing already that needs correction. You usually add two paragraphs between two classes. You made a big gap here in line 3916 and 3917. This should be avoided for readability.

https://github.com/Bforartists/Bforartists/pull/902/commits/58387124febc286e51aeeff33a41fcb176d8380c#diff-6382b06ba650d03ff16d0df4b906873e

Draise14:
Alright, so.. only do one task then pull for one each sitting or so. I removed the empty the spaces you mentioned and recommitted.

ReinerBforartists:
Thanks :)

Ah, i see you added new lines to the menus in the c files. Could you please next time add a comment there? A hint that this change comes from us? Something like:

/*bfa - modified menu, added spaces*/

Background is, changes in c can easily get lost with merges. I don't have all changes in my head. And maybe we stumble across it in a year or two again. When it's an icon or a tooltip, then the change is obvious. But such new lines are not so obvious.

ReinerBforartists:
Merged

Issue #924, opened 2019-05-31
Video Sequencer - Add Effect sub menu has missing icons, causes errors


Draise14:
## Issue
The Add Submenu for Effects is now broken, pulling errors. There are referenced icons to some that were instanced to a redundant node icon. Some of these icons are cleaned up and no-longer exist or are in another area due to this task #912 .  

![bforartists_2019-05-31_09-31-47](https://user-images.githubusercontent.com/35357409/58712904-1b19a580-8387-11e9-9eb5-e14e87b94b16.png)

## Setup
BFA2 Alpha 0.4.0, build 5/30/2019, Windows 10

## Task
The trouble icons in the **space_sequencer.py** file need to be updated to the cleaned up icons from task #912

PR #925, opened 2019-05-31
Video Sequencer - Add Effect sub menu has missing icons, causes errors #924

Draise14:
- Fixed line 591 with alternative icon


ReinerBforartists:
The commas to separate hot air is just a Blender "we do it our way" thingie to which everyone crazily agreed. We are not bound to it. But it doesn't hurt neither since it doesn't break the code. So that's just a side comment ...

ReinerBforartists:
Merged

Issue #1000, opened 2019-07-03
Video Sequencer - compare merge side panel changes from Blender 2.8 to Bforartists 2


Draise14:
# Issue
Blender: 
![blender_2019-07-03_11-02-47](https://user-images.githubusercontent.com/35357409/60607367-bfc24500-9d82-11e9-8663-4274394fa224.png)
Bforartist 2: 
![bforartists_2019-07-03_11-02-59](https://user-images.githubusercontent.com/35357409/60607212-7a9e1300-9d82-11e9-9d61-c1c1d6bc1cf8.png)

Seems there has been a regression that is pulling in errors and doesn't have some of the latest UI updates with menus/submenus from the 2.8 paradigm. We have 139 codes missing or in conflict from the new updated master of Beta Blender 2.8.0. 

![atom_2019-07-03_11-05-04](https://user-images.githubusercontent.com/35357409/60607253-8db0e300-9d82-11e9-9221-fe953ec4cf29.png)

# Task
Go through the code and merge in the correct code preserving BFA changes.

ReinerBforartists:
Merges -.-

Draise14:
Ok, done, pull request.  Took a bit of effort but now I can edit video like normal with normal features.

Issue #998, opened 2019-07-03
Video Sequencer - Strip Type not defined error


Draise14:
# Issue
![bforartists_2019-07-02_17-16-56](https://user-images.githubusercontent.com/35357409/60605421-1b8acf00-9d7f-11e9-918c-9f8fcea7d0ee.png)

There are some operator classes not defined correctly. 

# Task
Go to the lines and fix things up. There may be an operator that no-longer exists in the API (comparing to Blender 2.8).

Draise14:
Fixed - but I found a bunch of new inconsistencies.......

PR #1001, opened 2019-07-03
Video Sequencer - Strip Type not defined error #998

Draise14:
- I commented out one of the conflicts at line 1220, as task #1000 may prove that there are more issues that may need to be dealt with or fixed naturally with the correct merge.
- I added the correct class to the original issue (missing).


ReinerBforartists:
Removing codeparts does not look like a valid solution here. How does the blender code part look here?

Draise14:
The entry doesn't exist for the sequencer in 2.8 now, I just Grepwined it. Hmm?

ReinerBforartists:
Okay, then let's merge it.

Issue #997, opened 2019-07-03
Video Sequencer - wrong icon error


Draise14:
![bforartists_2019-07-03_10-33-31](https://user-images.githubusercontent.com/35357409/60605263-d1095280-9d7e-11e9-981e-7714897d4c66.png)
# Issue
Copydown icon is wrong in the **video_sequencer.py** file. Throws a console error. 

# Task
Fix it.

PR #1002, opened 2019-07-03
Video Sequencer - compare merge side panel changes from Blender 2.8 t…

Draise14:
Video Sequencer - compare merge side panel changes from Blender 2.8 to Bforartists 2 #1000

- Merged in most of the 2.8 changes
- kept in the header refresh button for now
- made sure all the BFA changes are kept and are working
- checked and squashed all bugs in the workflow, but not with a heavy video edit test

I know this isn't standard, but I also fixed this task in the process to clear out the console issues:
Video Sequencer - wrong icon error #997


ReinerBforartists:
Done

Issue #903, opened 2019-05-21
Video Sequencer - Strip Menu needs missing icons


Draise14:
![image](https://user-images.githubusercontent.com/35357409/58118614-dd6a8f00-7bc6-11e9-8ce0-80137b4a58a9.png)

## Task
Add missing icons to: 

- Snap Strips to Playhead
- Clear Strip Offset
- Toggle Meta

Draise14:
![inkscape_2019-05-30_17-34-24](https://user-images.githubusercontent.com/35357409/58669254-2e822d80-8301-11e9-8812-8e6e5edf53ef.png)

Ok, made an icon. Gonna unassign incase anyone wanna do icons. But DIBS, this one is mine.

Draise14:
![inkscape_2019-05-31_09-15-55](https://user-images.githubusercontent.com/35357409/58711915-210e8700-8385-11e9-8b84-61797fc305bc.png)
 
A little more progress. I have one more icon to work on then I'll generate icons, do the code and commit.

Draise14:
Gonna pause this one till task #924 is merged.

Draise14:
DONE! 

- Added 3 new icons
- assigned some reused ones to similar functionality

Draise14:
![bforartists_2019-07-03_10-45-55](https://user-images.githubusercontent.com/35357409/60605771-be434d80-9d7f-11e9-9651-7cc5261e78ac.png)

# Task
Re-add icons that have gone missing.

ReinerBforartists:
Sorry ^^

PR #1007, opened 2019-07-05
Video Sequencer - Strip Menu needs missing icons #903

Draise14:
- added back a bunch of missing icons both to context menu and header menu.


ReinerBforartists:
Done

Issue #1538, opened 2020-05-12
Preferences - Add Open Preferences Folder to Preferences save dialogues


Draise14:
# Issue
When developing addons, fixing addon bugs, and also when you need to do a hard factory reset for testing.. I've noticed a lot of users don't know how to get to the appdata config folder to troubleshoot, and/or I go there way to often to be comfortable fixing things. 

A user in the discord threw an idea with some code, and I scripted it a little more to make it hopefully relative: 

Tadaa! 

An entry to open the appdata folder: 
![image](https://user-images.githubusercontent.com/35357409/81711325-2a2f2180-9439-11ea-80b3-ee33fcec9d01.png)

# Task
Add an operator with a redirect icon to open the relative configuration path. 

### Reason
It will show users that a configuration folder exists when you use BFA through the UI (in Blender it's just.... expected prior knowledge), and with that exposed, users can also potentially easily learn how to factory reset to troubleshoot if needed, or get to know where their addons/script/hotkeys/configurations are. It is just to show and give access to artists and developers without any prior knowledge of how to get to these settings. It will also make it easier to give support when testing bugs when people sometimes request help in Facebook, Discord or in the forums. 

### Notes
I've already implemented and use this now. It's handy. 
I have a commit for this. 
I don't know if the relative path will work in MAC or Linux. This needs testing.

ReinerBforartists:
Looks indeed useful. I need to think this through. 

I will test if it works under Linux when i find the time for.

ReinerBforartists:
First thought: the preferences should be at the end of the menu. They are most used, and that way they are easiest to reach. The folder is something that a user opens once in a lifetime. Which raises the second question. Wouldn't this menu item not better fit into the preferences panel?

Draise14:
Good point. I put it on a surface level due to the script editor, for loading scripts for addons and things, but if it's psychologically more related to preferences, then yes.. it should go in the preferences, not in the Edit category. 

![image](https://user-images.githubusercontent.com/35357409/81715886-58633000-943e-11ea-80ca-aece00bbe68d.png)

Sidenote: this menu has missing icons? 

**Observations:** 
Maybe we can add "Open Preferences Folder" into the preferences editor.
Then also add "Open Scripts Folder" into the script editor, which would be the subfolder for the addons/scripts.

ReinerBforartists:
Yes, yes and yes. Please make a tracker entry ^^

ReinerBforartists:
[...]

PR #1540, opened 2020-05-12
Preferences - Add Open Preferences Folder to Preferences save dialogues #1538

Draise14:
This had a chunk of messy commits, but now it's ok. I have tested it by delete/adding the appdata folder, and things are conditionally going to LOCAL or USER accordingly. 

The messy commits was due to leaving in comments, first conditional was not working, and the first commit for the idea wasn't quite right. 

This may need testing in Linux and Mac to make sure it's ok, but according to:
[This link](https://docs.blender.org/manual/en/2.79/getting_started/installing/configuration/directories.html), it shoooould be relative to the OS according to how Blender works internally with paths. 

### Note
The USER goes to parent folder, showing all bforartists preferences. Though this might not be necessary.

### DIFF
--- a/release/scripts/startup/bl_ui/space_userpref.py
+++ b/release/scripts/startup/bl_ui/space_userpref.py
@@ -18,6 +18,17 @@
 
 # <pep8 compliant>
 import bpy
+
+#work
+import os 
+from pathlib import Path
+
+user_path = Path(bpy.utils.resource_path('USER')).parent
+local_path = Path(bpy.utils.resource_path('LOCAL')).parent
+
+
+
+
 from bpy.types import (
     Header,
     Menu,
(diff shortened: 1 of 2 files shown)


release/scripts/startup/bl_ui/space_userpref.py
```diff
@@ -133,15 +136,15 @@ def draw(self, context):
         layout.operator_context = 'INVOKE_AREA'
         layout.operator("wm.read_factory_userpref", text="Load Factory Preferences")
 
-#work
+#work        
         layout.separator()
 
-        if os.path.isdir(Path(bpy.utils.resource_path('USER'))):
+        if os.path.isdir(Path(bpy.utils.resource_path('USER'))): 
             layout.operator("wm.path_open", text="Open Appdata Folder", icon = "FOLDER_REDIRECT").filepath = str(user_path)
```
Draise14, line 143: "Open Preference Folder" might make more sense.

ReinerBforartists:
Not so fast with the pull requests. First finish the relevant tasks ^^

Do we have the same menu at two locations now?

And what about Linux and Mac? Have you tested it?

Draise14:
It's in one location. The old version in the first proposal was reverted. 

I don't have a linux build on hand, and my mac is decomissioned right now. I'm pending someone to test I suppose. I guess this testing process should be done in the task before pulling?

Draise14:
I think I'll close this for now, till I've tested it. I've found a couple of linux users who can try out the script for me. Looking for a mac user now.

PR #1743, opened 2020-08-13
Add a button to enable/disable volumetric blending

lordloki:
Basically, when active the new checkbox it fixes the ugly volumetric banding.

### DIFF
--- a/release/scripts/startup/bl_ui/properties_render.py
+++ b/release/scripts/startup/bl_ui/properties_render.py
@@ -299,6 +299,18 @@ class RENDER_PT_eevee_volumetric_shadows(RenderButtonsPanel, Panel):
 
         layout.active = props.use_volumetric_shadows
         layout.prop(props, "volumetric_shadow_samples", text="Samples")
+ 
+ class RENDER_PT_eevee_volumetric_blending(RenderButtonsPanel, Panel):
+    bl_label = "Volumetric Blending"
+    bl_parent_id = "RENDER_PT_eevee_volumetric"
+    COMPAT_ENGINES = {'BLENDER_EEVEE'}
+
+    def draw_header(self, context):
+        scene = context.scene
+        props = scene.eevee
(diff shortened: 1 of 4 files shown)


lordloki:
Yes, the "fix" is for eevee only. Specifically, for eevee viewport as eevee render has not this issue.

The button location is under Volumetrics tab in render 
image

Here you have a video showing volumetric banding (before) and volumetric blending (after). This video is for another fork (UPBGE) but you can see the volumetric banding issue and how the volumetric is blend "correctly"after.
https://www.youtube.com/watch?v=onCc3O4P_S8

Regarding to the implementation, Volumetric's Eevee is based in a Frostbite paper. This implementation blends the 5% of the new frame. This is the algorithm used for Eevee render BUT for Eevee viewport no blend is performed and therefore a ugly banding raises when you move through the viewport and a volumetric is used.

ReinerBforartists:
Nice improvement :)

May i ask for one tiny change? Could you please be so kind and mark the changes with a comment so that i don't accidentally remove it with the next merge? :)

ReinerBforartists:
Ah, nevermind, i'll do that.

ReinerBforartists:
Ahaha, merging, closing, and now i notice something. Wouldn't it make more sense to have the tab in the viewport settings? It's viewport only :)

Anyways, thanks for the commit :)

ReinerBforartists:
And another point. The checkbox is now a subtab without content. Hmmm. I should not accept pull requests so fast i guess XD

That's a new issue for our tracker i would say. Just wanted to mention it :)

checkboxissubpanel

lordloki:
Yes, in your case maybe the viewport setting would be more appropiated.
In our case (UPBGE) we had in Volumetric tab as they are the setting that affects to Game Engine "render".
I leave it up to you :-)

ReinerBforartists:
I see. Thanks. We will adopt it to our needs :)

Issue #1745, opened 2020-08-14
Linux deb - wrong startericon


ReinerBforartists:
We have still a problem with our Linux installation. When i start Bforartists, then the icon in the sidebar is the one from Blender.

![startericonbug](https://user-images.githubusercontent.com/13919702/90233374-584c4500-de1e-11ea-9aaf-44393a47a5f3.jpg)

ReinerBforartists:
Ah, forgot to mention, it is not only the icon. When you hover over the icon then it shows Blender as the name ...

![blender](https://user-images.githubusercontent.com/13919702/90234440-f68cda80-de1f-11ea-8f7e-6dd26ad7736b.png)

ReinerBforartists:
Testing ...

ReinerBforartists:
Houston we have a problem ^^

![broken](https://user-images.githubusercontent.com/13919702/90306942-652a7080-ded2-11ea-97b4-8de2673fd93d.jpg)

ReinerBforartists:
Clean build solved this issue. But the starter icon is still wrong.

![still](https://user-images.githubusercontent.com/13919702/90307498-eab01f80-ded6-11ea-9bcd-7e7d6920a22c.jpg)

ReinerBforartists:
Possible fix. The binary version starts now with the correct icon.

ReinerBforartists:
This issue is solved. Closing.

PR #1801, opened 2020-08-28
Add release files for Darwin (Mac) build.

seanr:
I'm not 100% certain that this has everything it _should_ have, but the builds do compile with this. I have one specific question that needs to be resolved about these, though:

Splash screen appears correct:
Screen Shot 2020-08-28 at 2 10 41 PM

About window appears to have Blender info rather than Bforartists:
Screen Shot 2020-08-28 at 2 11 04 PM

@ReinerBforartists I need some direction at least on where that about window is getting its info. How is this being done for the Windows and Linux versions?


ReinerBforartists:
Hmm. Are you sure that you need to add new files?

Blender builds for mac without any new files. And this means Bforartists should already have all required files onboard. Just with a few wrong brandings here and there.

I will have a look about the location of the files tomorrow. It is more than one.

ReinerBforartists:
Ah, the splash screen window mentions Blender because we want the user to know at which Blender version the current Bforartists version is based on.

*Bforartists 2.3.0 **is based on** Blender 2.90 Alpha*

The splash image shows Bforartists as it should. This info string appears in Windows and Linux the same way. And should also appear in a Mac build this way.

The string is defined in release\scripts\startup\bl_operators\wm.py , line 2579

Just ignore. All is well with that :)

ReinerBforartists:
Okay, question resolved. You say it compiles with your changes. 

Need your okay and if you have further questions. Then i will merge the changes :)

ReinerBforartists (approved):

seanr:
> Hmm. Are you sure that you need to add new files?
> 
> Blender builds for mac without any new files. And this means Bforartists should already have all required files onboard. Just with a few wrong brandings here and there.

It'll build named as blender.app, not bforartists.app and you'll end up with this weird hybrid. We need our build to use our specific stuff (icons, properties list, etc) and have the correct filename. I suspect the previous person doing the builds did this but didn't add it to the repo, though I'd certainly be interested in hearing exactly how they did it if you're still in contact with them.

> The splash image shows Bforartists as it should. This info string appears in Windows and Linux the same way. And should also appear in a Mac build this way.

The screenshots got flipped somehow. The splash screen is correct indeed, it's the about window that's wrong.

ReinerBforartists:
Ah, i see. Thanks for explanation. Just wanted to make sure that there is nothing unnecessary coming into the master All good  :)

Our last mac maintainer tried to sell Bforartists in a shop. And we lost contact then since he didn't react to any mail anymore. 

> The screenshots got flipped somehow. The splash screen is correct indeed, it's the about window that's wrong.

As told, when it's about the string in the splash screen, this string is intended. But i think you mean something different. Do you mean the icon in the taskbar and the tooltip for it?

How old is your branch? have fixed a icon and name problem yesterday at Linux. Possible that this fixes your problem too.

https://github.com/Bforartists/Bforartists/issues/1745

The commit:  https://github.com/Bforartists/Bforartists/commit/6c86680c8bbc77aaea87e9415d75ed0a76bc84ea

ReinerBforartists:
Merging this changes now. We can fix the icon at a later point.

seanr:
@ReinerBforartists just to clean up loose ends in this thread here: my branch was very fresh, from a matter of a half hour or so before the PR, IIRC. Regarding your other question, I mean this:

about widow

I think that's a Mac-only thing as it's provided by the operating system, but it gets the data from somewhere in the application and as you can see, that is currently incorrect.

ReinerBforartists:
Ah, now i see. It pulls the Blender version number, not the Bforartists version number. I will have a look if i can catch this somewhere.

My problem is, i cannot test it, i don't have a mac. But we will see.

Many thanks for your effort so far :)

PR #1889, opened 2020-10-21
task #1813, required projected falloff shape to tooltip of front faces only 

### LINKED ISSUE #1813: Weight Painting - Tools tab - Brushes panel - add hint to required projected falloff shape to tooltip of front faces only
Front faces only just works with falloff shape projected. With the default falloff shape sphere it doesn't work. Users might consider this tool as broken therefore.

Add a hint to the tooltip that this tool requires the falloff shape projected.

Maybe we can even grey out or hide the option as long as the falloff shape is sphere. This is one of the rare cases where greyed out would make sense so that the user knows this tool exists.

frontfaces

iyadahmed:
Fixes #1813 
new pop up message
"Brush only affects vertexes that face the viewer. Projected falloff only"
it also greys out option if falloff option is Sphere
image
image

### DIFF
--- a/release/scripts/startup/bl_ui/properties_paint_common.py
+++ b/release/scripts/startup/bl_ui/properties_paint_common.py
@@ -906,7 +906,9 @@ def brush_shared_settings(layout, context, brush, popover=False):
         layout.separator()
         
     if use_frontface:
-        layout.prop(brush, "use_frontface", text="Front Faces Only")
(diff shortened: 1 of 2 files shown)


ReinerBforartists:
Hm, wasn't the job to put a prop at another location? :)

Please! Always make a tracker issue first before you change something at the code!

PR #1891, opened 2020-10-21
Sculpt Mode - Tool tab - Brush Settings - Align Front Faces Only prop left

### LINKED ISSUE #1890: Sculpt Mode - Tool tab - Brush Settings - Align Front Faces Only prop left
We align props at the left side.

Removed part, since we agree about the change to be a good thing:
(And we don't grey out if we don't have to. We don't show the prop then.

There are corner cases where it makes sense to show a greyed out prop. And this prop here could be one of them. So we should discuss this part.)

alignpropleft

iyadahmed:
Fixes #1890 
image

### DIFF
--- a/release/scripts/startup/bl_ui/properties_paint_common.py
+++ b/release/scripts/startup/bl_ui/properties_paint_common.py
@@ -906,11 +906,10 @@ def brush_shared_settings(layout, context, brush, popover=False):
         layout.separator()
         
     if use_frontface:
-        layout.use_property_split = False
         col = layout.column()
+        col.use_property_split = False
         col.prop(brush, "use_frontface", text="Front Faces Only")
         col.active = False if brush.falloff_shape == "SPHERE" else True
-        layout.use_property_split = True
         layout.separator()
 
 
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Sorry. that was propably caused by my comment. In this case you could also use col.use_property_split = False. And it is in fact the better solution. With col just the current column is affected. And there is also no need to set it to True then again. This is always something to test if a property_split_false leads to unexpected side effects below the current change.

Could you please replace layout.use_property_split = False by col.use_property_split = False? And remove the layout.use_property_split = True?

iyadahmed:
> Sorry. that was propably caused by my comment. In this case you could also use col.use_property_split = False. And it is in fact the better solution. With col just the current column is affected. And there is also no need to set it to True then again. This is always something to test if a property_split_false leads to unexpected side effects below the current change.
> 
> Could you please replace layout.use_property_split = False by col.use_property_split = False? And remove the layout.use_property_split = True?

sure will replace it now

ReinerBforartists:
Perfect. Many thanks :)

iyadahmed:
Done

PR #1900, opened 2020-10-23
Hide the autokeyframe settings instead of greying out

### LINKED ISSUE #1895: Timeline - Hide the autokeyframe settings instead of greying out.
These autokeyframe settings are new. And greyed out. Visual noise. You don't need these settings as long as auto keyframing is not on.

hide

iyadahmed:
Fixes #1895 
image


ReinerBforartists:
Yep, it's really that simple :)

PR #1910, opened 2020-10-24
UV Editor - Options - Pixel Snap Mode improvements #1905

### LINKED ISSUE #1905: UV Editor - Options - Pixel Snap Mode improvements
The current implementation is a bit too noisy for my flavour. Neither the icons nor the text is really good to read.

I see two options here. Either we use pure icon buttons. Or we use pure text buttons.

What speaks for the pure icon buttons is that we could make it to a one liner, which reduces visual noise. What speaks for the text buttons is that you don't need to hover over the button to read the tooltip ...

And i don't like the term pixel snap mode. A snap to pixels is much clearer imho.

Suggestions? Ideas?

pixelsnapmode

iyadahmed:
Fixes #1905 
image

### DIFF
--- a/release/scripts/startup/bl_ui/space_image.py
+++ b/release/scripts/startup/bl_ui/space_image.py
@@ -1053,10 +1053,8 @@ class IMAGE_PT_image_options(Panel):
         col.prop(sima, "use_realtime_update")
         col.prop(uv, "show_metadata")
 
-        col = layout.column()
-        col.label(text = "Pixel Snap Mode")
-        row = col.row()
-        row.prop(uv, "pixel_snap_mode", expand = True)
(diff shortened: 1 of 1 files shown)


iyadahmed:
will resolve conflict because of the previous commit "hide snap to pixels in image editor"

ReinerBforartists:
It's a trap!

iyadahmed:
oh

iyadahmed:
I deleted the if statement for check if space data mode is UV
maybe that's the reason
adding it back

iyadahmed:
will also do a pull rebase from master

iyadahmed:
I just resolved the conflict on my branch, all is fine :)

Issue #1892, opened 2020-10-22
3D View - Hotkey only Operator - Transform Axis Target on lights has no UI entry


Draise14:
# Issue
The SHIT+T operator "transform Axis to Target" when selecting a light has no UI entry, not in the Right Click specials menu nor in the light context sensitive menu. it's a hotkey only operator. 

`bpy.ops.object.transform_axis_target()`

![image](https://user-images.githubusercontent.com/35357409/96886667-6c797780-1449-11eb-80ce-11135efb158f.png)

# Task
Needs an icon
Needs to be placed both in the context based light selection operators and double right click Specials menu

ReinerBforartists:
I can have a look, but as far as i remember it's a mouse operator.

Draise14:
Even so.. never knew it existed. Could we still add it to the UI so it's easier to manage it's hotkey and know that the feature exists? With a tooltip to say it's an under mouse operation. 

Maybe we can show these kinds of mouse only operators in the UI with the tooltip and a little asterisk in the name saying it's there to know it exists, learn about it, understand it - but it's use is hotkey recommended. 

Like this:
**Transform Axis to Target***


_I literally have used Blender and sometimes BFA for years before finding out the hotkey exclusive operators - and one of the reasons I use BFA is because you've added many to the UI again._

ReinerBforartists:
As told, i can have a look. But no promises. 

I have a section for the hotkey only functionality in every editor chapter in the first chapter of it. And i think this entry is listed there. If not then we need to add it there.

ReinerBforartists:
Hm, the context menu is mouse position dependant. Here we could definitely add the operator.

ReinerBforartists:
Now what to do with the regular UI entry. The object menu would be a place for it, when a lamp of the correct type is selected. We would need to add a hint towards hotkey only functionality then.

iyadahmed:
hmmm, should we edit the hint message?
![image](https://user-images.githubusercontent.com/25552173/97203812-fe70df80-17bd-11eb-93d4-468c40d4b182.png)

ReinerBforartists:
Hm hm hm, never tested with a camera object. But yeah, does not work.

The functionality would be of interest for the camera too though. That's the point where one would have to start some investigation in the Blender manual ^^

iyadahmed:
indeed, the camera doesn't seem to work

maybe I should check the source of the operator
[...]

PR #1926, opened 2020-10-26
3D View - Hotkey only Operator - Transform Axis Target on lights has …

iyadahmed:
…no UI entry #1892
image

### DIFF
--- a/release/scripts/startup/bl_ui/space_view3d.py
+++ b/release/scripts/startup/bl_ui/space_view3d.py
@@ -3044,6 +3044,9 @@ class VIEW3D_MT_object_context_menu(Menu):
                 props.data_path_item = "data.spot_blend"
                 props.input_scale = -0.01
                 props.header_text = "Spot Blend: %.2f"
+            
+            if light.type in ['SPOT', 'SUN', 'AREA']:
+                props = layout.operator("object.transform_axis_target", text="Transform Axis to Target", icon = "NODE_LIGHTPATH")
 
             layout.separator()
 
--- a/release/scripts/startup/bl_ui/space_view3d.py
+++ b/release/scripts/startup/bl_ui/space_view3d.py
@@ -2801,7 +2801,7 @@ class VIEW3D_MT_object(Menu):
                 props.header_text = "Spot Blend: %.2f"
             
             if light.type in ['SPOT', 'SUN', 'AREA']:
-                props = layout.operator("object.transform_axis_target", text="Track to Cursor", icon = "NODE_LIGHTPATH")
+                props = layout.operator("object.transform_axis_target", text="Interactive Light Track", icon = "NODE_LIGHTPATH")
 
 class VIEW3D_MT_object_convert(Menu):
     bl_label = "Convert To"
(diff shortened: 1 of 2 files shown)


ReinerBforartists:
As told in the issue, we should add the operator to the object menu too. And update the tooltip.

iyadahmed:
> As told in the issue, we should add the operator to the object menu too. And update the tooltip.

oh ok will do now

iyadahmed:
while looking at operator source, edited tool tip, and found a nicer name for the menu entry
image

iyadahmed:
restored original tool-tip, renamed menu entry
image

ReinerBforartists:
Great so far. Remains to remove the hint to the camera and to add the hint for the useage of the tool.

iyadahmed:
adding now

ReinerBforartists:
Don't forget to rename the entry in the object menu too.

iyadahmed:
> Don't forget to rename the entry in the object menu too.

already did :)

iyadahmed:
wrote this usage hint, how does it read to you?
image

ReinerBforartists:
We're getting close! But it is not clear with your tooltip that you first need to select the target object, then shift select the light , and then you can perform the tool. The object and shift select part is missing. Please change it to the following:

Select object, shift select light so that both are selected, perform tool

iyadahmed:
ok, will do now, the reason is I couldn't make it work by selecting target, it seems to make no difference if target is not selected, I might be missing something, changing now.

ReinerBforartists:
When you use it without an object selected then the light just rotates around in screen space. With a target object selected it will point in direction of the object in screen space, and will point through this object.

It's an awkward tool,  no question ^^

iyadahmed:
> When you use it without an object selected then the light just rotates around in screen space. With a target object selected it will point in direction of the object in screen space, and will point through this object.
> 
> It's an awkward tool, no question ^^

oh I see, indeed ^^

ReinerBforartists:
Yeah, it's hard to see. When you use two views then it is better noticeable. Took me also a while to figure it out back in the days when i first found out about this tool ^^

iyadahmed:
> Yeah, it's hard to see. When you use two views then it is better noticeable. Took me also a while to figure it out back in the days when i first found out about this tool ^^

oh makes sense ^^

iyadahmed:
Updated message
      "Interactively point lights to a location (Ctrl translates and aligns to target normal)\n"
      "(optionally select a target Object and a Light, then use this operator)"

added "optionally" because it can be used without, if not suitable can be changed

ReinerBforartists:
> wrote this usage hint, how does it read to you?
> image

It is still wrong XD
And while at it, it also misses the hint towards hotkey tool.

Interactively point a light towards an object (Ctrl translates and aligns to target normal)
Select object and shift select light. Perform tool and point to target
Hotkey tool! Please use the hotkey!

I would suggest to use exact this text so that we can finally close this pull request :)

iyadahmed:
> > wrote this usage hint, how does it read to you?
> > image
> 
> It is still wrong XD
> And while at it, it also misses the hint towards hotkey tool.
> 
> Interactively point a light towards an object (Ctrl translates and aligns to target normal)
> Select object and shift select light. Perform tool and point to target
> Without target object the light rotates freely
> Hotkey tool! Please use the hotkey!
> 
> I would suggest to use exact this text so that we can finally close this pull request :)
> 
> Note to myself. Yet another proof why it is better to split the tasks into as small steps as possible. And not to extend a task more and more. We should have split this up into its parts. Next time ...

ok will do now, I arrived home late yesterday, just woke up now

iyadahmed:
Done
image

ReinerBforartists:
Aaand closed :)

PR #1931, opened 2020-10-28
Issue 1920

### LINKED ISSUE #1920: Properties Editor -  Physics - Fluid - Liquid subtab - hide greyed out props for Fractional obstacles
... , add a triangle button at the end, and indent the Obstacle Distance and the Threshold prop to mark it as sub content of the Fractional Obstacles checkbox.

We hide, we don't grey out when we don't have to.

hideindent

iyadahmed:
Fixes #1920 
image


ReinerBforartists:
Looks good. But could you please indent the sub content? :)

PR #1944, opened 2020-10-30
Transform panel in Edit mode - text inside sliders #1940

### LINKED ISSUE #1940: Transform panel in Edit mode - text inside sliders
In Edit mode with a primitive object we have some sliders that has still the text of the prop inside of the silders. This is inconsistent with the rest of the UI. Text with the name of the prop should be in front of the sliders. 

Mesh object, curve object, surface object, metaball object, bone object, lattice object.

transformedit

iyadahmed:
Fixes #1940 
image

### DIFF
--- a/source/blender/editors/space_view3d/view3d_buttons.c
+++ b/source/blender/editors/space_view3d/view3d_buttons.c
@@ -212,6 +212,8 @@ static TransformProperties *v3d_transform_props_ensure(View3D *v3d)
 /* is used for both read and write... */
 static void v3d_editvertex_buts(uiLayout *layout, View3D *v3d, Object *ob, float lim)
 {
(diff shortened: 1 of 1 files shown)


source/blender/editors/space_view3d/view3d_buttons.c
```diff
@@ -435,33 +437,41 @@ static void v3d_editvertex_buts(uiLayout *layout, View3D *v3d, Object *ob, float
     const int butw = 200;
     const int buth = 20 * UI_DPI_FAC;
     const int but_margin = 2;
-    const char *c;
```
iyadahmed, line 438: not needed any more, text is being used directly, not stored the used

PR #1951, opened 2020-10-31
Transform panel - text inside sliders #1940 - edit bone

iyadahmed:
is there a better arrangement for this, should there be more separators?
image


ReinerBforartists:
Could indeed be a bit more separated. Hm.

ReinerBforartists:
A useful separator gap would be beween the Tail Z and the radius prop i would say.

iyadahmed:
added a separator after radius, radius is linked to tail X/Y/Z (tail radius), just like head radius
image

PR #2019, opened 2020-11-26
NLA Editor - Sidebar - make View panel content more compact

### LINKED ISSUE #1629: NLA Editor - Sidebar - make View panel content more compact
The panel items could stand much compacter to each other. And there is a big gap at the top ...

nlasidebar

iyadahmed:
Fixes #1629

### DIFF
--- a/release/scripts/startup/bl_ui/space_nla.py
+++ b/release/scripts/startup/bl_ui/space_nla.py
@@ -184,7 +184,6 @@ class NLA_PT_view_view_options(Panel):
 
         st = context.space_data
 
-        layout.separator()
 
         layout.prop(st, "show_markers")
         layout.prop(st, "show_local_markers")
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Ah, my description of the task was not clear enough. Removing the layout.separator() is just half of the task. When we use (align = True) then we can put the props even closer together.  Which looks nicer then. 

Would you be so kind to have a look at this too? Else i would need to open yet another task for it  :)

iyadahmed:
> Ah, my description of the task was not clear enough. Removing the layout.separator() is just half of the task. When we use (align = True) then we can put the props even closer together. Which looks nicer then.
> 
> Would you be so kind to have a look at this too? Else i would need to open yet another task for it :)

Sure, will do just saw the comment

ReinerBforartists:
Many thanks :)

PR #2020, opened 2020-11-26
Timeline/Dope sheet/NLA editors - slider covers button to call sidebar

### LINKED ISSUE #1867: Timeline/Dope sheet/NLA editors - slider covers button to call sidebar
The slider covers the button to call the sidebar here.

I could swear i had this reported already. But it is not to find. Anyways, Blender has no content in the sidebar anymore in the dope sheet editor. We have. So it's our responsibility to fix this bug i guess.

dopesheetslidercoverssidebar

iyadahmed:
Fixes #1867 
note: as of now doesn't handle old files (2.90 and less), but correctly handle default files (startup.blend)
image


ReinerBforartists:
Ah templates remains a can of worms. 

We have some trouble here. I have a 1920 x 1280 monitor. And here it shows some gaps now with our toolbar editor. Nothing that you could fix, it's a limit in the Blender UI engine, i had it reported before. So to change the layouts is something i have to do.

Just curious, how did you manage to add the compositing, masking and motion tracking workspaces to the video editing template? ^^

problem

ReinerBforartists:
What was the reason again that you needed to change the templates? I reinstalled the old startup.blends and the slider issue is still fixed. The buttons are not longer covered by the sliders.

iyadahmed:
yeah it was fixed because I changed versioning_defaults.c, this handles opening startup.blends (including templates), the file still has the issue, but once it is opened the property changes, so I just opened all defaults one by one and did save startup, then copied them over to the source

iyadahmed:
> Ah templates remains a can of worms.
> 
> We have some trouble here. I have a 1920 x 1280 monitor. And here it shows some gaps now with our toolbar editor. Nothing that you could fix, it's a limit in the Blender UI engine, i had it reported before. So to change the layouts is something i have to do.
> 
> Just curious, how did you manage to add the compositing, masking and motion tracking workspaces to the video editing template? ^^

I didn't change anything though, I just opened the template and saved startupfile, so adding compositing, masking and motion tracking was by accident ^^

ReinerBforartists:
Yeah, i told you it's a can of worms ^^

I think it is okay that some old files still shows the bad behaviour. The to break the startup blends is the bigger danger here.

iyadahmed:
> Yeah, i told you it's a can of worms ^^

Indeed
> 
> I think it is okay that some old files still shows the bad behaviour. To break the startup blends is the bigger danger here.

I think so too

Issue #1293, opened 2019-12-04
The tooltips on tabs show two lines instead of one line. 


raindropsfromsky:
Reporting for Bforartists 2 alpha 080, running on Windows 10 64 bit Version 10.0.18362 Build 18362.

Hover the mouse on any one of these tabs:
![image](https://user-images.githubusercontent.com/9047168/70169859-a6d30c80-16f1-11ea-996f-96ebc1cf3a3d.png)

A tooltip will appear. It shows two lines. [BUG]
The top line is blank, except only one period.
![image](https://user-images.githubusercontent.com/9047168/70169790-82773000-16f1-11ea-875b-ac74b9d12baf.png)

**Desired:** It should show only one tooltip line.

ReinerBforartists:
Thanks for the report. Let's have a look what we can do here :)

ReinerBforartists:
Related: Rework Tooltips #1031

ReinerBforartists:
Hm, not this easy. Now i know why i have left the dots alone. We either display the tooltips with their headers everywere. And this shows some hiccups like here. Or we do it the Blender way. Then i have to readd the tool name to a few thousand tooltips by hand again. Hm hm hm.

ReinerBforartists:
This problem is in more than one place.

![lightprefs](https://user-images.githubusercontent.com/13919702/70306757-f99ee800-1807-11ea-9023-adca9c3f77ea.png)

ReinerBforartists:
Fixed in master

I would have loved to add a description to those buttons too. But the current Blender code design does not allow this. I was able to keep our tooltips with the tool names intact, and remove the empty string with the dot at the end in the Properties tab buttons. 

Blender does not show the "Plane" Label at the button yet, and at many other buttons. There you have to guess the tool name from the button name. But we have them as icon buttons too. As told, it would have been several thousand tooltips again without a tool name in the tooltip.

![fixed](https://user-images.githubusercontent.com/13919702/70312697-73d56980-1814-11ea-911b-ad2fdedb9b6f.jpg)

PR #2074, opened 2020-12-07
Compilation warning - missing format arguments for BLI_sprintfN #2064

### LINKED ISSUE #2064: Compilation warning - missing format arguments for BLI_sprintfN
image

iyadahmed:
Fixes #2064 
image

### DIFF
--- a/source/blender/editors/interface/interface_region_tooltip.c
+++ b/source/blender/editors/interface/interface_region_tooltip.c
@@ -806,7 +806,7 @@ static uiTooltipData *ui_tooltip_data_from_button(bContext *C, uiBut *but)
     /*It just shows a dot then where the title should be. So we check for those buttons, and skip adding the button title*/
     if (STRPREFIX(but->drawstr, but_label.strinfo))
       {
-      field->text = BLI_sprintfN(but_label.strinfo);
+      field->text = BLI_sprintfN("%s", but_label.strinfo);
     }
     else if (!STRPREFIX(but->drawstr, but_label.strinfo)) {
       field->text = BLI_sprintfN("%s.", but_label.strinfo);


ReinerBforartists:
I will download the branch and make a few tests first. I still fear that you break the functionality with your change back. There are some very special tooltip titles that doesn't show when you check for but_label.strinfo.

The culprit here is this check, wich i have disabled for good reason: if (but_label.strinfo && !STRPREFIX(but->drawstr, but_label.strinfo)) {

iyadahmed:
> I will download the branch and make a few tests first. I still fear that you break the functionality with your change back. There are some very special tooltip titles that doesn't show when you check for but_label.strinfo.
> 
> The culprit here is this check, wich i have disabled for good reason: if (but_label.strinfo && !STRPREFIX(but->drawstr, but_label.strinfo)) {

alright, will also try to solve the warning without killing the feature now

iyadahmed:
I think simply using %s without . in the first condition will solve the warning, without changing what the code does

ReinerBforartists:
Great. And sorry for the misunderstandings. I am not this good at working at several things at once :)

The mean part is that the tooltips and the titles are implemented in the usual Blender manner. Everything in its own way. Plus the design decision not to show the tool name in the title everywhere. I have never understood the reason why they let the user guess what tool or menu it is.

iyadahmed:
Indeed, I think it is sorted out now, also we can just remove the first condition, but will use an empty string as format string, because using %s makes an empty space at the top of the tooltip, but let's test first, which tooltips are affected by this?

ReinerBforartists:
> I think simply using "%s" without . in the first condition will solve the warning, without changing what the code does

Isn't that the dot at the end? Ah, my C fu is simply too weak ^^

iyadahmed:
> > I think simply using "%s" without . in the first condition will solve the warning, without changing what the code does
> 
> Isn't that the dot at the end? Ah, my C fu is simply too weak ^^

ah yep we remove the dot, that what cause the visual issue, but if we leave the %s it simply displays an empty string (value but_label.strinfo), which takes space at the top of the tooltip

iyadahmed:
I think this would fix, but not sure if it affects the specific tooltips, need to test
image

ReinerBforartists:
> Indeed I complicated things ^^, I'm sorry I made multiple changes, it is not good that way,
> I think it is sorted out now, also we can just remove the first condition, but will use an empty string as format string, because using %s makes an empty space at the top of the tooltip, but let's test first, which tooltips are affected by this?

It is more than a dozen from what i remember. I would have to collect them again too. The range goes across simple operators across menu names , property names and so on. Some works as thought, some not. I don't find the original issue at the moment. My poor hands ... :(

But affected are for example the buttons to create the primitives in our toolbar at the top.

You could add a test string and go through the UI to see what is affected and what not.

iyadahmed:
> > Indeed I complicated things ^^, I'm sorry I made multiple changes, it is not good that way,
> > I think it is sorted out now, also we can just remove the first condition, but will use an empty string as format string, because using %s makes an empty space at the top of the tooltip, but let's test first, which tooltips are affected by this?
> 
> No harm done. That's development ^^
> 
> It is more than a dozen from what i remember. I would have to collect them again too. The range goes across simple operators across menu names , property names and so on. Some works as thought, some not. I don't find the original issue at the moment. My poor hands ... :(
> 
> But affected are for example the buttons to create the primitives in our toolbar at the top.
> 
> You could add a test string and go through the UI to see what is affected and what not.

right ^^, ok great will test now

iyadahmed:
it crashes if we remove it entirely, so trying the other way

ReinerBforartists:
Ah you already found the original issue it seems. https://github.com/Bforartists/Bforartists/issues/1293

But i haven't listed all affected elements here. The problem of this task was that the tabs in the properties editor did show a dot above the tooltip. And here one line is enough. No title required.

iyadahmed:
will upload the branch again, I was concerned about an empty space at the top of some tooltips, but i turned out it was there already, the change would be to use "%s" in the first condition too, no difference after change, but removed the warning

ReinerBforartists:
When we have unexpected spaces in the tooltip titles then this would be another task. Do you have an example? :)

iyadahmed:
sure the tooltips that had a dot before, now has an empty space
image

ReinerBforartists:
Ah, i see. Never found a solution for that. You can make a task for it if you want.

iyadahmed:
great making a task now

PR #2159, opened 2021-01-10
New expanded property UI style #2033

### LINKED ISSUE #2033: New expanded property UI style
I'm not happy about how much space the labels steals here from the edit boxes. Just not sure how to fix this one here.

One solution would be to put the label at the top above instead in the first x row, then the x row x with just an X label , closely followed by the edit box. This would give us most space here i guess.

morespace

Mockup: 

mockup

iyadahmed:
Fixes #2033 
New style right, old left
Note: this changes for all expanded prop with use_property_split set to True
some props are done "manually", like Dimensions in the screenshot, that's for another task
image

### DIFF
--- a/source/blender/editors/interface/interface_layout.c
+++ b/source/blender/editors/interface/interface_layout.c
@@ -1992,6 +1992,9 @@ void uiItemFullR(uiLayout *layout,
   bool use_prop_sep_split_label = use_prop_sep;
   bool use_split_empty_name = (flag & UI_ITEM_R_SPLIT_EMPTY_NAME);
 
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Phew. Looks pretty complicated ^^

Merging. The fine tuning will follow in other tasks then :)

iyadahmed:
> 
> 
> Phew. Looks pretty complicated ^^
> 
> Merging. The fine tuning will follow in other tasks then :)

it isn't complicated after finding the right combination ^^

great, looking into these

PR #2167, opened 2021-01-12
Modifier header, switch remove modifier button with the menu #2129

### LINKED ISSUE #2129: Modifier header, switch remove modifier button with the menu
For my flavour the x button to close the modifier is a bit too close to the apply button. This could lead to accidental clicks. So imho it would be better when the delete button is at the end. The menu acts then as a separator to the apply button. Which reduces missclicks. And the delete button should be at the end of the list anyways.

Task, switch positions of delete modifier button and the modifier header menu.

switchxandmenu

iyadahmed:
Fixes #2129
image


ReinerBforartists:
Love it :)

Issue #2043, opened 2020-12-02
Node Editors - iconize add menus


ReinerBforartists:
We have an add menu in every node editor. I was never able to add an icon here, the content of this menus is an array, which is then shown as a menu. And i have never found a way to connect icons to it.

The node arrays can be found in \release\scripts\startup\nodeitems_builtins.py
The NODE_MT_add menu is in Release\2.92\scripts\startup\bl_ui\space_node.py
And the def draw_node_categories_menu is in release\scripts\modules\nodeitems_utils.py

@iyadahmed , do you maybe know a way to connect the arrays with icons so that we can iconize the add menus here too?

![addmenu](https://user-images.githubusercontent.com/13919702/100846768-3505de00-347f-11eb-8551-b794ca4ca32d.jpg)

iyadahmed:
Investigating into it after finishing the other task

iyadahmed:
![image](https://user-images.githubusercontent.com/25552173/105544361-9588fd00-5d03-11eb-99a8-fa6c962a2264.png)
it turned out this is as easy as setting icon of layout operator in release/scripts/modules/nodeitems_utils.py
we could define a dictionary to map from node type to icon, what would be better is to somehow make use of the node header icons we added earlier #1609
![image](https://user-images.githubusercontent.com/25552173/105544527-c10be780-5d03-11eb-8f9a-4afde743f50e.png)

iyadahmed:
Progress

![image](https://user-images.githubusercontent.com/25552173/108002216-90b02400-6ff7-11eb-97e5-7145994414ee.png)
![Screenshot 2021-02-16 004218](https://user-images.githubusercontent.com/25552173/108002225-9443ab00-6ff7-11eb-895a-6878fb8e2995.png)

ReinerBforartists:
Nice =3

PR #2168, opened 2021-01-15
Node Editor - Expose listed Node Groups dynamically in Node shelf. #2151

### LINKED ISSUE #2151: Node Editor - Expose listed Node Groups dynamically in Node shelf.
# Issue
image

Currently, node groups are only accessible in the Add menu then group sub-menu. You can't just click and drag them into use. You have to search, or go into sub-menus to use them. 

# Feature Request
Bforartists exposes nodes into a panel, it would be great to be able to have these group nodes also exposed in the side panel with names dynamically, so it's easier to see what groups are available and to re-use them easily.

iyadahmed:
Fixes #2151 
Note: this is shared between all node editors
a task to be done is renaming "Input" tab in Texture/Comp/Shader nodes toolshelf to "Add"
so that it matches Geometry nodes one

image

the code is borrowed from 
release/scripts/modules/nodeitems_utils.py
release/scripts/startup/nodeitems_builtins.py

P.S: we could use the exact menu code directly, but it would show the Make Group and Ungroup buttons
and also won't be able to have icons, unless we edited the menu code, so that gave more clues to #2043

### DIFF
--- a/release/scripts/startup/bl_ui/space_node_toolshelf.py
+++ b/release/scripts/startup/bl_ui/space_node_toolshelf.py
@@ -1593,6 +1593,60 @@ class NODES_PT_Input_output_tex(bpy.types.Panel):
             props.type = "TextureNodeViewer"
 
 
+#Input nodes tab, Group panel
+from nodeitems_builtins import node_tree_group_type
+
+# from nodeitems_builtin, not directly importable
+def contains_group(nodetree, group):
+    if nodetree == group:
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
The Input category should remain Input category, because we have also other categories here. I have divided it into three parts since in the other editors it became too much to display it in one single add panel. The Geometry Node Editor is different here. It has just one tab called Add.

PR #2223, opened 2021-02-05
Motion Tracking - sliders with text in the slider #1866

### LINKED ISSUE #1866: Motion Tracking - sliders with text in the slider
The whole tracking workspace contains sliders that has the description text inside of the slider instead in front or above it. Inconsistent. Either we have in the whole UI the prop name in front or above the sliders, or inside of the sliders. Even when it eats more space then.

slidertext

iyadahmed:
Fixes #1866 

The new style is great, but I'm a little worried about this inconsistency in label column width, should we discuss this?
image

### DIFF
--- a/source/blender/editors/space_clip/clip_buttons.c
+++ b/source/blender/editors/space_clip/clip_buttons.c
@@ -539,30 +539,34 @@ void uiTemplateMarker(uiLayout *layout,
                  0,
                  tip);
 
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Hm. I think we can live with this little inconsistency. You have solved it in a good way :)

iyadahmed:
Ok great :)

PR #2242, opened 2021-02-15
Node Editors - iconize add menus #2043

### LINKED ISSUE #2043: Node Editors - iconize add menus

iyadahmed:
Fixes #2043 
Added optional argument "icon" to NodeItem
did the first few nodes

image

Screenshot 2021-02-16 004218


ReinerBforartists:
Do you start from scratch? ^^

iyadahmed:
> Do you start from scratch? ^^

Not entirely, but made it in a much better way I believe #2308

PR #2255, opened 2021-02-22
Animation - Apply Animation to all selected through UI Animate proper…

### LINKED ISSUE #2163: Animation - Apply Animation to all selected through UI Animate property buttons
# Issue
bforartists_i2dNYBJh37

Currently, when you animate with characters or multiple objects, you can key all of them by pressing I then selecting which properties you want to animate from a very long un-iconized list. 

Like in Bforartsits as in Blender, if you use the UI entries, you can only set animation on Active object properties.
bforartists_vb1Zrgy5dx

In general, the UX for all selected is holding down ALT then editing the sliders and it would apply to all selected. 

# Feature Request
It would improve the UX if we could animate all selected by holding down ALT then pressing the keyframe UI entry, then it would add and set keyframes for all selected, just like the "I" then Keyframe Panel operators. 

This may be... a global change, but here the focus is to focus on the animation properties of the Transform Panel

# Task
image

- Research to see if this is possible, to be able to use the UI entry to then trigger animation key to all selected
- Evaluate how this can be implemented
- Evaluate if this is a feature that should be committed to Blender directly, or set here

## Method
I've asked @iyadahmed to do some research into this, document it and find away to save some time figuring it out to give us some options.

iyadahmed:
Fixes #2163 
…ty buttons #2163

### DIFF
--- a/source/blender/editors/animation/keyframing.c
+++ b/source/blender/editors/animation/keyframing.c
@@ -2438,6 +2438,7 @@ static int insert_key_button_exec(bContext *C, wmOperator *op)
   const AnimationEvalContext anim_eval_context = BKE_animsys_eval_context_construct(
       CTX_data_depsgraph_pointer(C), (float)CFRA);
   bool changed = false;
+  bool changed_any = false;
   int index;
   const bool all = RNA_boolean_get(op->ptr, "all");
   eInsertKeyFlags flag = INSERTKEY_NOFLAGS;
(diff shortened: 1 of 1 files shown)


Draise14:
Just tested this out, the apply animation to bones is currently not fully working, so this needs more work before being pulled. So.. needs a bit more work.

iyadahmed:
Yep It is currently a draft

ReinerBforartists:
And drafts cannot be merged. So please be patient ;)

I have a permanent eye at the pull requests. But currently i have not this much to add to the two drafts. @iyadahmed you can always poke me in case i overlook something.

iyadahmed:
Alright :)

ReinerBforartists:
Hey Iyad,

Shall we finally close this request until you are ready to commit? The master evolves permanently ... :)

iyadahmed:
> Hey Iyad,
> 
> 
> 
> Shall we finally close this request until you are ready to commit? The master evolves permanently ... :)

Sure :), we can close for now, until the patch becomes ready

PR #2273, opened 2021-03-02
Updated Blender merge (fixes OSL in macOS builds)

seanr:
The merge from upstream Blender just before 2.8.0 missed a crucial update to the Mac cmake script resulting in a partially broken build - the build on the site now works fine except that it is missing OSL. I have a dev build now, built off this branch, that contains the fix for that. We need to make sure this gets merged down so it's in our next release. The most relevant upstream commit is this one:

https://github.com/Bforartists/Bforartists/commit/e00a87163cfd1b885afb4cb3ddff5f9fe911d2aa

Additionally, I'd like to get the merge documentation into the Wiki here (@Draise14 set me a Word doc with instruction 😱). We don't really currently have build instructions other than a single page linking to PDFs and that should really live in a place actually attached to the repo itself, ideally the Wiki tab right here in Github - the Readme can then reference that if we want.


ReinerBforartists:
Sure. So you need write access to the wiki?

ReinerBforartists:
Ah, as a sidenote, please always push to the master if even possible. The sync_blender_master branch is to maintain the Blender changes.

seanr:
The Blender commits were the merge from Blender into this repo - I followed the same procedure in the word doc that Draise sent me describing the BFA merge process. Since it's merged now, I think we're all good. I've been using it for nearly a week in a Mac dev build with no issues. :)

Issue #2044, opened 2020-12-02
Node Editors - tooltips for the single nodes


ReinerBforartists:
The add menu in the node editors does not have proper tooltips for the single nodes so far.The tooltips all says "Add a node to the active tree". A short description of what the node does would do miracles here.

I failed at this one before, but that's no reason not to try it again ...

![tooltip](https://user-images.githubusercontent.com/13919702/100849539-ff62f400-3482-11eb-9799-c5a529603080.jpg)

iyadahmed:
since 2.81 It is possible to dynamically change operator tooltip by defining "description" classmethod
found answer here
https://blender.stackexchange.com/questions/56727/changing-tooltip-for-menu-entries-which-use-the-same-operator
which mentioned this link
https://wiki.blender.org/wiki/Reference/Release_Notes/2.81/Python_API#Operators

by modifying release/scripts/startup/bl_operators/node.py
![image](https://user-images.githubusercontent.com/25552173/105554776-b441c080-5d10-11eb-8e8a-35a024ddd577.png)

result 
![image](https://user-images.githubusercontent.com/25552173/105554671-7e9cd780-5d10-11eb-9132-333c4830b159.png)

ReinerBforartists:
@iyadahmed , shall we go the def description route here? I remember you mentioned somewhere that the tooltips are already defined somewhere, but are simply blank for now.

ReinerBforartists:
Ah, found it. https://github.com/Bforartists/Bforartists/pull/2308

> PS: NOD_static_types.h already has a place for description for each node which we can use for this task #2044
> description can be then accessed in NodeItem just like label and icon
> bpy.types.Node.bl_rna_get_subclass(self.nodetype).description

ReinerBforartists:
Hm, i am lost. I have added the def description to the node.py as shown. But i have no idea where i should add our node descriptions then.

Halp :(

ReinerBforartists:
This should work, shouldn't it? But throws me an error that NODE_OT_add_node has no attribute 'arg'

```
#Input nodes tab, connect panel. Just in shader mode
class NODES_PT_Input_connect(bpy.types.Panel):
    """Creates a Panel in the Object properties window"""
    bl_label = "Connect"
    bl_space_type = 'NODE_EDITOR'
    bl_region_type = 'UI'
    bl_category = "Input"
    
    arg: bpy.props.StringProperty()
    
    @classmethod
    def description(cls, context, properties):
[...]

PR #2308, opened 2021-03-24
Issue 2043: Node add menu icons

### LINKED ISSUE #2043: Node Editors - iconize add menus

iyadahmed:
Fixes #2043 
image

This patch sets icons for node types so it can be defined once, and accessed in C and Python
I also moved the ugly switch case from node_draw.c :), and soon shall be removed entirely
 
this is how to get icon from node type after this patch
`bpy.types.Node.bl_rna_get_subclass(nodetype).icon`
nodetype is for example: `"ShaderNodeAmbientOcclusion"`
which the patch uses in node add menu icons

a similar thing was already used for node labels by Blender
`bpy.types.Node.bl_rna_get_subclass(nodetype).name`

Blender however didn't expose icon, nor did it set icons for nodes, so I had to modify some C files, it wasn't clear at first, but it was worth it at the end

once the icon is set for each type (I reused the switch case we wrote before, for now), and icon property exposed to Python
we can access icons in C or Python
in C we can do this
`RNA_struct_ui_icon(node->typeinfo->rna_ext.srna)`
which after this patch is used in node headers icons

TODO: define icons in source\blender\nodes\NOD_static_types.h instead of the switch

PS: NOD_static_types.h already has a place for description for each node which we can use for this task #2044 
description can be then accessed in NodeItem just like label and icon
`bpy.types.Node.bl_rna_get_subclass(self.nodetype).description`
image


ReinerBforartists:
I don't fully understand the code. But it works. So i am happy :)

ReinerBforartists:
Before i forget, super sexy. Excellent work :)

iyadahmed:
> Before i forget, super sexy. Excellent work :)

Thank you 😄

PR #2317, opened 2021-03-26
Node icon implementation problems - expected a string enum, not dict …

### LINKED ISSUE #2316: Node icon implementation problems - expected a string enum, not dict
Houston we have a problem with our new icons in the add menu of the node editors.

The new argument icon leads to icons in the add menu. But also to a problem now. The new nodegroups fails to show in the group menu.

Not sure if this is a general problem, or connected to my latest change where i tried to give this node groups an icon. Investigating ...

bug

blendergroup

iyadahmed:
…#2316
Fixes #2316

image

### DIFF
--- a/release/scripts/modules/nodeitems_utils.py
+++ b/release/scripts/modules/nodeitems_utils.py
@@ -43,7 +43,7 @@ class NodeCategory:
 
 
 class NodeItem:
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Many thanks for the fix :)

PR #2366, opened 2021-04-18
Merge Blender 3.0.0

iyadahmed:


ReinerBforartists:
Hey iyad, 

Looks great so far, but i'm a bit surprised to see another branch here. Seems that i wasn't good enough in explaining, my mistake. 

It is not good to create an extra branch for the merge. The sync_blender_master branch is especially meant for just the merges. And the merges must happen in this branch, since this branch gives us a fallback option. Which is gone when we create new branches for every merge and delete them afterwards.

The pipeline is as told, switch to sync_blender_master, merge and resolve the issues from the merge. Commit to sync_blender_master. This part should work for you without the need to make a pull request first. Then do a pull request to merge the sync_blender_master branch into the master. No third branch involved :)

So could you please merge your changes into the sync_blender_master branch first? And from there then into master? :)

iyadahmed:
Ah ok will do now

ReinerBforartists:
I notice differences between my merge and your merge. Have you done this step here?

> Now compare the following folders with the Blender repo. That's the sub modules. And they don't update automatically, we don't use the Blender sub modules.
> 
> Don't copy over the git, gitignore and arcconfig files!
> 
> release\scripts\addons - DON'T blindly copy over. We have more than one Bforartists change here, and own scripts. Some im- and exporters have icons. And we have modified the materials_library_vx addon.
> release\scripts\addons_contrib - no content from us. So it is save to copy the files over in case there are changes.
> release\datafiles\locale - The language files
> source\tools - I have modified the blender_theme_as_c.py here, and added a comment how to use it. Rest should be equal to Blender
>

iyadahmed:
> I notice differences between my merge and your merge. Have you done this step here?
> 
> > Now compare the following folders with the Blender repo. That's the sub modules. And they don't update automatically, we don't use the Blender sub modules.
> > Don't copy over the git, gitignore and arcconfig files!
> > **release\scripts\addons** - DON'T blindly copy over. We have more than one Bforartists change here, and own scripts. Some im- and exporters have icons. And we have modified the materials_library_vx addon.
> > **release\scripts\addons_contrib** - no content from us. So it is save to copy the files over in case there are changes.
> > **release\datafiles\locale** - The language files
> > **source\tools** - I have modified the blender_theme_as_c.py here, and added a comment how to use it. Rest should be equal to Blender

hmm, I did, where is the differences

iyadahmed:
ah it is in locale
one second will fix

ReinerBforartists:
In 
release\scripts\addons
release\scripts\addons_contrib
release\datafiles\locale 
and
source\tools

:D



differences

iyadahmed:
hmm ok, I did copy the folders, made sure  addons and addons contrib has recent changes, maybe I did something wrong, nut yeah I did the pull yesterday

iyadahmed:
ah wait let me copy these again

ReinerBforartists:
Just to make sure, the image above shows the differences between your merge and my merge. Not between Blender and Bfa. Ideally there shouldn't be any differences since we both did the same merge. But obviously there are some. That's why i asked :)

I have also merged yesterday. Usually there are no pushes at sunday to the Blender repo. That's why i prefer sunday as the merge day. It's not this dramatic when something doesn't fit here. There is always the next merge ...

iyadahmed:
alright working on it at the moment :)

ReinerBforartists:
No hurry. Exact is better than fast ;)

iyadahmed:
I copied locale

addons seems to be correct, hmmm
I compared BFA to the Blender master I used
image

ReinerBforartists:
Looks good :)

source\tools ?

iyadahmed:
> Looks good :)
> 
> source\tools ?

checking now

iyadahmed:
I also just made sure source tools is copied and pushed it
here's the compare now

ReinerBforartists:
Looks good too :)

Let me pull the latest pull request and compare it with what i have here ...

ReinerBforartists:
Let's continue in the new pull request ...

PR #2409, opened 2021-05-09
Issue 2404

### LINKED ISSUE #2404: Missing Category warnings
To resurrect the tabs has some unwanted side effects. The console is now flooded with missing category warnings. We should do something here.

Either we supress the warning. Or we update all the panel classes ...

warnings

iyadahmed:
Fixes #2404 
the issue was a bit deeper, lesson learned, test both Debug and Release


iyadahmed:
> @iyadahmed , looks good :)
> 
> But could you please add a comment to the change in source/blender/makesrna/intern/rna_ui.c ? Else i might kill this change with the next merge :)
> 
> I usually deactivate old code in the c files instead of removing it. That way you can follow what was removed, and the comment then tells why.

Sure, doing so now

PR #2412, opened 2021-05-10
Node Editors - tooltips for the single nodes #2044

### LINKED ISSUE #2044: Node Editors - tooltips for the single nodes

iyadahmed:
Fixes #2044 

Now we should fill descriptions for the rest of Nodes in source/blender/nodes/NOD_static_types.h

Screenshot 2021-05-10 222136

### DIFF
--- a/release/scripts/modules/nodeitems_utils.py
+++ b/release/scripts/modules/nodeitems_utils.py
@@ -43,7 +43,7 @@ class NodeCategory:
 
 
 class NodeItem:
-    def __init__(self, nodetype, label=None, settings=None, poll=None, icon='NONE'):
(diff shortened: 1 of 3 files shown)


ReinerBforartists:
I would have never found this one. Thanks a bunch :)

Hm, the tooltip is still missing in the sidebar with icons only. And this is what this task is mainly about. With icon only you don't have any hint what the tool is. Only the python expression gives us a hint at the moment.

Could you please have a look if you can fix this part too?

missing

iyadahmed:
Uploading a new pull request

PR #2447, opened 2021-05-22
Node Editors - tooltips for the single nodes #2044

### LINKED ISSUE #2044: Node Editors - tooltips for the single nodes

iyadahmed:
Fixes #2044 
image


ReinerBforartists:
Super sexy!!!

Many thanks for this commit :)

PR #2474, opened 2021-06-03
Transparent background for the toolshelf tools #2398

### LINKED ISSUE #2398: Transparent background for the toolshelf tools
Not sure if manageable independently. But at the moment it looks odd with the grey background. Maybe we can make just the tools panel background transparent, but keep the background of the other panels intact.

tools

iyadahmed:
Fixes #2398 
This adds a new possible value to bl_options, "HIDE_BG", it sets alpha for panel background to 0
This PR also replaces HIDE_HEADER with HIDE_BG for toolsystem panels
image

### DIFF
--- a/release/scripts/startup/bl_ui/space_toolsystem_toolbar.py
+++ b/release/scripts/startup/bl_ui/space_toolsystem_toolbar.py
@@ -2526,7 +2526,7 @@ class IMAGE_PT_tools_active(ToolSelectPanelHelper, Panel):
     bl_region_type = 'TOOLS'
     bl_label = "Tools"
     bl_category = "Tools"
-    bl_options = {'HIDE_BG'}
+    bl_options = {'HIDE_HEADER'}
 
     # Satisfy the 'ToolSelectPanelHelper' API.
     keymap_prefix = "Image Editor Tool:"
@@ -2618,7 +2618,7 @@ class NODE_PT_tools_active(ToolSelectPanelHelper, Panel):
     bl_region_type = 'TOOLS'
     bl_label = "Tools"
     bl_category = "Tools"
-    bl_options = {'HIDE_BG'}
+    bl_options = {'HIDE_HEADER'}
 
     # Satisfy the 'ToolSelectPanelHelper' API.
     keymap_prefix = "Node Editor Tool:"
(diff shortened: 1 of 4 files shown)


ReinerBforartists:
Sexy :)

Mh, i would love to keep the original tools with the HIDE_HEADER opion on, not with the new option to hide just the background. That way the old and new tools are easier to distinguish. The panels with header are the old tools then. And sorting is still possible. You simply sort the old tools around the new ones. And when you turn off the tabs then you have the exact look like in Blender. No header, transparent background.

iyadahmed:
> Sexy :)
> 
> 
> 
> Mh, i would love to keep the original tools with the HIDE_HEADER opion on, not with the new option to hide just the background. That way the old and new tools are easier to distinguish. The panels with header are the old tools then. And sorting is still possible. You simply sort the old tools around the new ones. And when you turn off the tabs then you have the exact look like in Blender. No header, transparent background.
> 
> 
> 
> EDIT, oh and could you please add our bfa comment to every change in the C code? As told, the danger is that i kill the change with a merge then when there is no proper comment that tells me it is our code.

Many thanks 🙏
Sure will edit it now

iyadahmed:
Hmm, but if we revert to HIDE_HEADER, then the new option is useless, not sure

ReinerBforartists:
As told, the new tool system should stay the old way. The hidden background is just meant for the old toolsystem.

iyadahmed:
Ah I see now, ok changing

Draise14:
> 
> 
> As told, the new tool system should stay the old way, with HIDE_HEADER. The hidden background is just meant for the old toolsystem, our new tabs, where we use HIDE_BG then, to hide the background, but keep the header visible.
> 
> We can always change the new toolsystem to show the toolbar there too. But my belly tells me that it is better to hide the header for this toolset. It doesn't only show in the 3d view ...

I would prefer the new tool system with a header, not only for consistency but it is easier to pin that shelf for all the tabs instead of all the panels from all the tabs to the toolshelf tab with the only group that can't be pinned. The pinning and re-ordering should go both ways for the shortest route possible, new and old.

ReinerBforartists:
And i would prefer not to have a header here:

notgood

And with tooltabs off:

off

As told, we can discuss this at a later point. It does not need to be solved now, in this task. This task here is just to enable the option at all.

iyadahmed:
Changes done ✅ 
As for the discussion, it is annoying to have header hidden,
I also agree that it should like Blender, but, maybe we could make it like that when tabs are hidden, but that's for another issue...

PR #2521, opened 2021-06-27
Text showing in toolbar tabs when it is not supposed to #2520

### LINKED ISSUE #2520: Text showing in toolbar tabs when it is not supposed to
image

iyadahmed:
fixing a small issue
Fixes #2520 
image


ReinerBforartists:
Good catch. Thanks for this :)

Issue #2565, opened 2021-07-27
3D View - Insert Keyframe menu icons


Draise14:
# Request
![image](https://user-images.githubusercontent.com/35357409/127234605-41bb3a65-ec5f-470b-acf5-5d1f9d277892.png)

The insert keyframe menu is large, and.... a wall of text. You press I to keyframe, and this is what you get. 

It would be nice if there were icons.

Draise14:
> I just had the idea of an array it needs to basically match the length of builtin keying sets
> and order of icons must match the order we need
> here's the mall having same icon

![image](https://user-images.githubusercontent.com/35357409/138311990-bcd30c07-bc0b-4e14-a9cc-1bfd02456744.png)

He's a genius. 

So I guess we have a technique that is not too intrusive in the code via if statements and enum list length, so less merge headaches. 

The other method is actually working the enum array with the icon, but that is harder, more invasive. 

Next question is.. what icons go where? Any new icons needed?

ReinerBforartists:
Give me the code and i will try to fill it with icons :)
A few already exists. Like the delta ones. And for loc rot scale we can simply use the move rotate and scale icons. But in this list it is not important that every tool has its own icon. But more a visual separation into groups. We could even get away with some colored keyframe icons here, which would make most sense to me. The other icons are too intrusive, with too much noise.

iyadahmed:
> Give me the code and i will try to fill it with icons :) A few already exists. Like the delta ones. And for loc rot scale we can simply use the move rotate and scale icons. But in this list it is not important that every tool has its own icon. But more a visual separation into groups. We could even get away with some colored keyframe icons here, which would make most sense to me. The other icons are too intrusive, with too much noise.

Sure one moment creating PR

ReinerBforartists:
Also here. Thanks for the pull request. But please be patient. I have just finished the devbuild before a few minutes. And tomorrow is merge. So it's a bad time to accept pull requests. They come in the way then.

I will accept the pull request when the merge is through. It is not forgotten, promised :)

ReinerBforartists:
Reopen since we need to add the icons here.

iyadahmed:
> So we need to add the icons here. Where is this done? in the int icons[] array?

yep

iyadahmed:
one moment let me push another pr

ReinerBforartists:
Fixed in master

The icons was quite a challenge ...

[...]

PR #2774, opened 2021-11-13
Set icons for builtin keying set enum items #2565

iyadahmed:


ReinerBforartists:
Thanks for the pull request. But please be patient. I have just finished the devbuild before a few minutes. And tomorrow is merge. So it's a bad time to accept pull requests. They come in the way then.

I will accept the pull request when the merge is through :)

Issue #2807, opened 2021-12-05
implicit declaration of function 'min'


ReinerBforartists:
Mnaether ran into a problem with the compiling of the release version of Bforartists at the centOS docker.

```
root/blender-git/Bforartists/source/blender/editors/animation/keyingsets.c:843:27: error: implicit declaration of function 'min'; did you mean 'fmin'? [-Werror=implicit-function-declaration]
843 |   int ks_icons_iter_len = min(ks_icons_len, ks_icons_idnames_len);
 |                           ^~~
|                           fmin
cc1: some warnings being treated as errors
```

The unfunny part is, it throws not even a warning here at Windows. It also compiles just fine at Ubuntu. So i am lost what the problem could be with this code.

@iyadahmed, this is your code here. The keyingset icons. Do you have an idea what is going on?

ReinerBforartists:
Related : 3D View - Insert Keyframe Menu - Icon order is misaligned in Armature>Pose Mode #2795
Related: 3D View - Insert Keyframe menu icons #2565

ReinerBforartists:
Correction, this problem also appears with Ubuntu.

![ubuntu](https://user-images.githubusercontent.com/13919702/144744663-0e420987-1f8c-458d-866d-4ead4302d060.jpg)

iyadahmed:
Hi will investigate asap

ReinerBforartists:
Thanks Iyad. Mnaether told me the nearly same solution. He has brackets in his term though. And i have already tested his solution at Windows and Ubuntu, and it works. 

I don't have the time to run this test twice. So i will use the method that works, and close your pull request. Nevertheless thanks for investigating :)

ReinerBforartists:
Pushed the fix also to the bforartists_v3.0.0 branch

PR #2808, opened 2021-12-05
Fix: implicit declaration of function 'min' #2807

iyadahmed:


ReinerBforartists:
As told, in the thread, solution was already at its work and tested when you made your pull request. Nevertheless thanks for it :)

iyadahmed:
Ok no problem:)

Issue #2978, opened 2022-04-29
Sequencer - Theming, New title bar might need some work (too black)


Draise14:
![bforartists_Ymwpmn275y](https://user-images.githubusercontent.com/35357409/165886543-21bfbc20-c222-4242-b6e5-a0a368a26c1b.png)
The new channel titles might need a better theming.

ReinerBforartists:
It's currently like in Blender ...

Draise14:
Been trying to dig deeper here, looks like this may be hardcoded.

ReinerBforartists:
:)

Draise14:
Found it:
https://developer.blender.org/rB277fa2f441f

ReinerBforartists:
Hm, userdef_default_theme.c means it is a theming option somewhere.

ReinerBforartists:
Welcome to development. Every task a new riddle and a new gamble :)

In such a case first try to find it in Blender. It is very likely that you haven't added these values to Bforartists userdef_default_theme.c yet. 

Note that you need to create a new userdef_default_theme.c file to have the new values in the factory theming. I think i have described this step before. You have to save out a userpref blend file. Which is usually automatically done anyways when you change anything at the preferences.

Then dive into the appdata directory, pick up the userpref.blend, and put this file into source\tools\utils

In this folder open blender_theme_as_c.py file. It contains further instructions. I hope they are clear enough :)

And afterwards, when the theming is fitting, you need to save out the Bforartists theme too. You will find it in the appdata dir then. And update it in the repo. These are two independent things.

And be super careful and triple check. This is one of the things where you can mix things up big time ...

![themeasc](https://user-images.githubusercontent.com/13919702/166499875-da75c8cf-ed93-4394-937c-8f5db5cf0a1a.jpg)

![advice](https://user-images.githubusercontent.com/13919702/166499662-cf907e7a-e3e6-4528-b490-eb287337079a.jpg)

![addtheme](https://user-images.githubusercontent.com/13919702/166500247-70386f65-4db9-4632-814a-56173a281639.jpg)

![path](https://user-images.githubusercontent.com/13919702/166500526-f0a0fe38-9dc1-4e02-8472-ce79ccc41097.jpg)

Draise14:
Huh, that's cool. That was easy enough, now compiling a test! Nice work on the comments and instructions! Saves on hand editing that C file.

Draise14:
[...]

PR #2991, opened 2022-05-03
#2978 Sequencer - Theming, New title bar might need some work (too black)

Draise14:
- Changed to lighter grey per the theme standards


ReinerBforartists:
As a reminder, you should not need to do a pull request anymore. Such things can go directly into the master.

I will merge it now.

Issue #2986, opened 2022-05-02
Geometry Nodes - New "Named Attribute" Nodes needs update


Draise14:
![rutview_XsEbYlR2Nv](https://user-images.githubusercontent.com/35357409/166331376-5e1e00e9-a9be-4546-9dcb-dfe646ad356b.png)

![rutview_WCBuMFQhMk](https://user-images.githubusercontent.com/35357409/166331379-896a1c6e-5b0f-4477-aec4-b952cbc0d86a.png)

Task:
- Add icons to new entries in the Add Menu's and Add shelves
- Update documentation
- Update tooltips

ReinerBforartists:
Also here, an image speaks thousand words ...

Draise14:
Images added

ReinerBforartists:
Do you want to try to create this two icons? :)

ReinerBforartists:
We have still the tutorial here: https://www.bforartists.de/data/tuts/Icontutorial.pdf

Note that i use the Visibone palette in Gimp. You should use this palette too to have consistent colors in the UI.

And don't forget to put the created icons also into our file repository. The png file, and the svg file. https://github.com/Bforartists/files/tree/master/iconset_bforartists_2

![visibonepalette](https://user-images.githubusercontent.com/13919702/166420911-eec08419-edfe-4308-9767-5513d79ee909.jpg)

Draise14:
I remember doing this once, but it is quite involved. I will have to try do this tomorrow. I already have a mind for the two icons:

- The attribute tag and delete together
- The attribute tag and save together

ReinerBforartists:
Don't forget to commit the created icons to the file repository when done :)

Draise14:
Gonna try give this a go now. I am trying to get two versions of Inkscape going, since I'm using the beta for speed and features - so 0.96 is a bit behind. Not sure how to have two versions of it though.

ReinerBforartists:
As a note, don't save the icon sheet with the new Inkscape version. The defaults are different. And our python blender_icons_update.py file differs from the blender_icons_update.py in Blender. You would screw up the icon creation.

Draise14:
Ok.. now compiling to test.

Draise14:
Eureka!!!! Back in the icon making business baby!

![image](https://user-images.githubusercontent.com/35357409/167237993-4b8ed4d5-d162-4f63-9133-9418db7bd72c.png)

Can close and commit now.

PR #2994, opened 2022-05-07
#2986 Added Icons to the Remove Attribute and Store Named Attribute

Draise14:
Will pull tomorrow after a good sleep.


ReinerBforartists:
This is a bug fix. this should go directly into the Master ;)

Merging ...

Issue #3238, opened 2022-09-18
Hot Keys - Node Editor - Select should be "Click" instead of "Press" so you can drag groups of nodes


Draise14:
![image](https://user-images.githubusercontent.com/35357409/190923089-888158af-63b5-42ca-8f1e-45518c19f91b.png)

Selecting nodes should be on click instead of press, so you can drag groups of nodes. Right now, when you have a group of nodes selected by box select, trying to click and drag them selects the current node or deselects the node - making dragging the group annoyingly impossible. 

I missed this one with my custom keymap, but the default has this issue.

Draise14:
Ah, this is a pretty big one. I'll commit this today.

PR #3268, opened 2022-10-07
Hot Keys - Node Editor - Select should be "Click" instead of "Press" …

Draise14:
…so you can drag groups of nodes #3238


Draise14:
Damn, forgot to pull the last commit

Issue #3360, opened 2022-12-23
Update space_toolbar.c in source\blender\editors\space_toolbar


ReinerBforartists:
We have our own toolbar editor. Sometimes Blender updates the structure for the editors, and adds and removes features from the editor files. 

I notice such a change that did not arrive in our editor. Decentralize .blend I/O for space types. I am not sure what effect this change will have. But when it doesn't hurt then i would love to include these changes into our editor type c file too. We might also have a look at the other two cleanup issues.

![blendwrite](https://user-images.githubusercontent.com/13919702/209322957-aa410914-cda8-4948-87e5-45309601f48c.jpg)

PR #3361, opened 2022-12-23
Cleanup space toolbar files to make them equal Blender

ReinerBforartists:
Changes regarding this task here: Update space_toolbar.c in source\blender\editors\space_toolbar #3360

@iyadahmed , would you please be so kind to have a look at these changes? I haven't found any negative side effects so far. But i am not sure if i introduce something breaking here.


iyadahmed:
I'm not sure too, but if it is equal to blender then good to go,
apologies for delay I just saw this

PR #3602, opened 2023-06-19
Fix for 3601-find-and-replace-in-text-editor-

Blenux:
- Added BFA Find and Replace Popup Panel Addon


Blenux:
currently cleaning up code

PR #3692, opened 2023-08-20
fix for 3687-3d-view---header---view---viewer-no

Blenux:
- Hidden the Viewer Node in the 3D View header menu, since one is already in the Overlays Options.


ReinerBforartists:
I think it would be better to completely remove the prop from the View menu. Our python files are this different from the C files that comments makes no sense. This case here is more or less self explaining since options always goes into an options panel in Bforartists. 

The only exceptions from this rule is the few options in the header of the View menu that cannot be placed in options really. Toggle sidebar and such. But even here i start to think from time to time what a better place could be ...

Blenux:
Ah true, will fix after taking kids to school today.

Blenux:
pushed fix2

PR #3777, opened 2023-09-18
Sync blender master

Draise14:


Draise14:
Mac, Ubuntu, fine. Trying out windows now, which works locally so will merge in now.

Issue #3834, opened 2023-10-03
Default Theme - Add header background transparency (for legibility)


Draise14:
![image](https://github.com/Bforartists/Bforartists/assets/35357409/ae928a8e-f76f-4432-81b9-ccc16eda7faf)

Alot of the header props are floating on.. nothing - with a lot going on it becomes hard to see what the rows are. They added the ability to make it.. kinda nicer. 

Let's make it at least the same transparency of the tabs backgrounds. This will make it easier to see the button rows, know it's a HUD for the 3D View and.. easier to read the light buttons and text on dark background.  

<details>
<summary>Details</summary>

SHA-1: 73460903fa94bc2302bbd39efdd8cc554447d642

* Asset Shelf: Transparent asset shelf header with background for buttons

Draw the background of the asset shelf header fully transparent, with an opaque
background with rounded corners behind sections containing buttons. This
reduces the visual space consumed by the asset shelf, and makes the header
follow a tabbed folder metaphor better. Also, this works much better with our
click-through feature, where transparent parts of regions without buttons are
passed through the region under it (we might want to consider unifying code
here a bit).

The edge to drag for region resizing respects the transparent sections.
When there is little space between sections, the sections get merged so that
there are no small gaps in the bar.

Part of #107881.

----

Note that the core of this is implemented in a generic way, so this can be
reused for other regions.

Pull Request: https://projects.blender.org/blender/blender/pulls/112241

</details>

<details>
<summary>Details</summary>

SHA-1: 55eaa755e3fcb0b9cb3417afee6f3e65b2c43d7b

* UI: Draw 3D view tools header transparent with background behind buttons

Makes use of the new buttons sections feature used by the asset shelf to
draw a background behind buttons, but the reset of the region fully
transparent. The UI team agreed on this as a way to minimize dead space,
but also as a better way to communicate our click through feature (where
dead space in overlapping regions forwards events to the underlying
region). The region background theme setting only affects the
non-transparent part of the region, the rest is always fully
transparent. With region overlap disabled, the bar is fully opaque still.

Also see #112241.

</details>

ReinerBforartists:
I am not really happy with making the header areas semi transparent again. We loose a part of the viewport area this way. I like the full transparency more.

Draise14:
[...]

PR #3846, opened 2023-10-05
Fix for 3844-theming-problem-in-preferences

Blenux:
-- adjusted the USERPREF_PT_save_preferences panel to match rest of Bforartists theme. 
-- adjusted the userdef_default_theme.c also value for above too (might be causing it?).

Screenshot_20231005_125656


ReinerBforartists:
Good job, but please also update the userpref.blend file in tools\utils

That the blend file is not updated shows me that you did edit the files manually. Which we should never do. I'm the same here with the Blender developers. Never edit the userdef_default_theme.c by hand. It needs to be created by the blender_theme_as_c.py script in the tools\utils directory. Else you might miss some entries.

The steps are described in the python scriptitself.

blenderthemeasc

Blenux:
Oh I didn't realize, my bad so do I revert the changes in this pull and only upload the .blend?

ReinerBforartists:
> Oh I didn't realize, my bad so do I revert the changes in this pull and only upload the .blend?

No, redo the whole procedure by doing it by script. As seen in the discord, we have new content from Blender. And this new content needs to arrive in our theming somehow. What you need to do is to follow all the steps, including regenerating the bforartists.xml file then from the preferences. As shown here:

https://github.com/Bforartists/Bforartists/wiki/Changing-the-Default-Theme

ReinerBforartists:
Looking at the changes, this looks good now. I'll merge it in, and make a new task for the other themes.

ReinerBforartists:
Related: https://github.com/Bforartists/Bforartists/issues/3834

PR #3847, opened 2023-10-05
fix for 3837-asset-browser-+-property-editor-modifiers-add-change-una…

Blenux:
-- changed Unassigned to Unassigned (Catalogue), and remove the Icon for consistency.


Draise14:
Looks good to me. If I see another instance of unassigned, I'll make a new task.

PR #3900, opened 2023-10-18
fix for 3891-geometry-nodes-sidebar-group-tab-interface-panel-same-to…

Blenux:
--- Separated add operators with own descriptions.

add_input
add_panel
add_output

### DIFF
--- a/scripts/startup/bl_operators/node.py
+++ b/scripts/startup/bl_operators/node.py
@@ -271,7 +271,7 @@ class NodeInterfaceOperator():
             return False
         return True
 
-
+# -bfa separated the add operators with their own descriptions.
 class NodeInterfaceOperator(Operator):
     def find_valid_socket_type(self, tree):
         socket_type = 'NodeSocketFloat'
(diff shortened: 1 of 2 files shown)


ReinerBforartists:
I know that we usually don't comment changes in the python files. But this mainly affects the files in startup/bl_ui. But you modified a python file in startup/bl_operators here. And there is a high danger that your change will be reverted with a merge then.

May i ask you for the favour to comment your changes in the node.py file? :)

Blenux:
Ok can do.

Blenux:
hows that?

iyadahmed:
Btw what do you mean poll not needed?

ReinerBforartists:
No idea, sorry, these commments are not from me. I think @iyadahmed Did these comments.

Mh, i meant this lines here. Every change from us should have a comment here.

thischanges

iyadahmed:
No worries all good ^^

Blenux:
ah right I will recheck and see if I can fix soon (totally forgot why I did it this way lol)

Blenux:
should be good now (fingers cross :smile: )

Issue #3780, opened 2023-09-18
Asset Browser - Display Mode - Thumbnails - Add button row for quick thumbnail sizes


Draise14:
![image](https://github.com/Bforartists/Bforartists/assets/35357409/48372fd1-d4c2-464b-a35e-0384bc2afc52)
New slider for thumbnails now possible, but the quick presets for standard views would be nice. 

- Tiny 48
- Small 96
- Medium 128
- Large 256

Previous options:
![image](https://github.com/Bforartists/Bforartists/assets/35357409/af42e09b-79ec-4730-9624-76130c0a4657)

Operator for the new buttons row
`bpy.context.space_data.params.display_size = 96`

Blenux:
Like so?

https://github.com/Bforartists/Bforartists/assets/25260650/9378288b-cc7c-4a3b-850a-2c9c0fb13e01

Draise14:
Yeah that's really nice!

Issue #519, opened 2019-01-04
Grease Pencil Object - Strokes menu in Draw Mode needs icons


ReinerBforartists:
Grease Pencil object is new, and its menus are missing the one or another icon.

Grease Pencil Object - Strokes menu in Draw Mode needs icons.

![satrokesindrawmode](https://user-images.githubusercontent.com/13919702/50684003-5ef4c280-1014-11e9-8081-8da4a5b321d6.jpg)

ReinerBforartists:
Duplicate. Already fixed.

![alreadyfixed](https://user-images.githubusercontent.com/13919702/53489986-c9653480-3a92-11e9-8698-19c06825625c.jpg)

PR #3907, opened 2023-10-31
fix for 3780-asset-browser-display-mode-thumbnails-add-button-row-for…

Blenux:
- added quick thumbnail sizes.
- adjust default sizes to match other areas to be consistent (menu drop downs from view header etc.

Screenshot_20231031_143525


ReinerBforartists:
May i remind to always update the manual with every commit? Or, if not possible, make a task in the manual tracker? These undocumented changes would make the manual unreliable.

I'll make a task ...

ReinerBforartists:
Created here: Document issue Asset Browser - Display Mode - Thumbnails - Add button row for quick thumbnail sizes #3780 #519

https://github.com/Bforartists/Manual/issues/519

PR #3928, opened 2023-11-07
fix for 3925-file-browser-add-thumbnail-button-sizes-to-keep-consiste…

Blenux:
- added Thumbnail Presets to Filebrowser.

image


ReinerBforartists:
Manual updated? ^^

PR #4052, opened 2024-01-27
Fixed Freedesktop Metainfo

StandingPadAnimations:
Renamed appstream references from `org.bforartists.Bforartists` to `de.bforartists.Bforartists`


ReinerBforartists:
Many thanks :)

Issue #4054, opened 2024-02-01
3D View - Object Mode - Animation - missing icon to clear active keying set


Draise14:
![bforartists_euqnFMd8JC](https://github.com/Bforartists/Bforartists/assets/35357409/b6b67549-bd59-4ccc-8572-350f0c03ee02)
Missing icon for the Set Active Keying Set floating panel

PR #4071, opened 2024-02-09
fix for #4054-3d-view-object-mode-animation-missing-icon-to-clear-acti…

Blenux:
-- added icon CLEAN_CHANNELS (CLEAN_KEYS might be another)

image


Blenux:
Reverted a commit, push to wrong branch.

PR #4082, opened 2024-02-17
fix for 4080-space-properties-tabs-separator-is-now-bigger-than-before

Blenux:
-- Fixed Spacing issue in the Space Properties Navigation Tabs


ReinerBforartists:
Does this code differ from Blender code? Then it would be wise to comment it so that the next merge does not kill it immediately again :)

Blenux:
Same has blender code.

PR #4103, opened 2024-02-22
fix 4102-ui-clean-up-purge-has-redundant-confirmation-dialogue

Blenux:
-- removed the outliner_orphans_purge_invoke operator, and it's related ops.


Draise14:
Looks good. I would comment out the code and comment it instead of removing it completely, to make merge work a bit easier.

Blenux:
ah fixed.

iyadahmed (changes requested):
Too much commented code even if removed instead of commented it would be too much removed code, I would vote for reducing the amount of changes to the minimum

ReinerBforartists:
> Too much commented code even if removed instead of commented it would be too much removed code, I would vote for reducing the amount of changes to the minimum

I think this looks fine in general. A bit confusing because of the C++ way to comment out every line with a double slash. The C++ way to use the backslash with stars is of course more elegant. And i think this is what Iyad means. I seem to rememeber that  the blender devs also explicitely asks for the C way to comment, and to avoid the C++ way.

C++ way:

```
//outcommented content 
//outcommented content 
//outcommented content 
```

C way:
```
\* 
outcommented content 
outcommented content 
outcommented content 
*\
```

Blenux:
I'll try to fix this today, once I get back home.

Blenux:
Anyone else wanna tackle this one, can't work out how to comment it out cleanly (without having used function errors during compiling) .

ReinerBforartists:
Hm, we have already a merge conflict here. New libs and stuff is also in the way. Would most probably be easier to start over i guess.

ReinerBforartists:
Excellent. Remains the merge conflict :)

PR #4132, opened 2024-03-10
fix 4129-timeline-mute-is-now-inversed-to-be-play-sound

Blenux:
-- Changed the Mute audio checkbox to off by default, so users don't wonder why strips audio ain't playing sound.
-- Updated Description to match checkbox.

### DIFF
--- a/scripts/startup/bl_ui/space_time.py
+++ b/scripts/startup/bl_ui/space_time.py
@@ -273,7 +273,7 @@ class TIME_PT_playback(TimelinePanelButtons, Panel):
         row.prop(scene, "use_audio_scrub", text="Scrubbing")
         row = col.row()
         row.separator()
-        row.prop(scene, "use_audio", text="Mute")
+        row.prop(scene, "use_audio", text="Mute", invert_checkbox=True)
 
         col = layout.column(align = True)
         col.label(text = "Playback")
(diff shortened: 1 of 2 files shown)


iyadahmed:
Sadly we should use invert checkbox instead
https://docs.blender.org/api/current/bpy.types.UILayout.html#bpy.types.UILayout.prop
image

iyadahmed (changes requested):
Sadly we should use invert checkbox instead
https://docs.blender.org/api/current/bpy.types.UILayout.html#bpy.types.UILayout.prop
image

Blenux:
bah overthinked it :grin:

Blenux:
updated

iyadahmed:
I'm sorry I suggested invert checkbox, invert checkbox is just a UI thing it doesn't make audio play by default, and also changing Blender behavior is not right either, the correct solution is to change the default value, this is what the intent is from what is written in the original PR,

so in short,
let's change the default value,
but not the logic nor invert checkbox nor change description, because changing description will make the description not describe what the option does in reality,
Let's please pause work on this, until we all figure out and agree on this

Blenux:
All good mate, will wait till when ready :smiley: 

How does one change the default value without changing the logic, though?

ReinerBforartists:
Wouldn't be the correct solution to adopt to the Blender solution? They changed it. And i don't see any benefit in keeping our solution different. Just more confusion at the user side.

iyadahmed:
> Wouldn't be the correct solution to adopt to the Blender solution? They changed it. And i don't see any benefit in keeping our solution different. Just more confusion at the user side.

This is right, and I think our code already matched Blender for this functionality, the pull request was going to make bfa different than Blender and probably confuse people, and also addons that rely on this property, and also loading Blender files

Draise14:
Gonna close this one and change the label with another commit. KISS

PR #4141, opened 2024-03-12
Test Branch for Topbar

Draise14:
Added in the topbar 2.0, thanks to @Blenux 

**New Features:**
- Now fixed to the header
- New preferences panel so a user doesn't have to zig zag to configure, and they can also reset to default settings 
- They can also use quick toggles on a per group type basis (instead of show all)
- Theme updated to use the toolbar nicely, including consistent workspace tab colours


Draise14:
Now waiting on merge to go to master.

Issue #4156, opened 2024-03-21
Workspaces - Add 3D Sequencer workspace for upcoming feature


Draise14:
![imagen](https://github.com/Bforartists/Bforartists/assets/35357409/cdf7048b-5347-4121-885d-d4ed6b636c05)

Add the workspace for the 3D Sequencer.

Draise14:
Now included in the Video Editing workspace group, and All.

Issue #4153, opened 2024-03-21
3D View - Sculpt Mode - Sculpt - New Lasso Show/Hide operators


Draise14:
![image](https://github.com/Bforartists/Bforartists/assets/35357409/7fc591ea-5cba-45a2-8b93-79c18a158092)
![image](https://github.com/Bforartists/Bforartists/assets/35357409/e8bc2575-6cd0-4db0-b398-90ea958e28bb)

New Lasso Show/Hide

May move this up and out for consistency. Will need two new icons. 

<details>
<summary>Details</summary>

SHA-1: 68afd225018e59d63431e02def51575bfc76f5cb

* Sculpt: Add Lasso Hide tool

This commit adds the `SCULPT_OT_hide_show_lasso_gesture` and the
corresponding Lasso Hide tool.

* Exposes the selection type for both the lasso and box hide tools
  as a option in the header
* Adds functionality into `sculpt_gesture.cc` for handling lasso
  selections with the `Outside` selection type

For `SelectionType::Outside`, the current implementation opts to not
do any filtering on the PBVH node level due to cases where the node
is mostly covered by a single gesture.

Addresses one of the tools in #80390

Pull Request: https://projects.blender.org/blender/blender/pulls/119140

</details>

Draise14:
![Uploading imagen.png…]()

Documenting...

Draise14:
Commit to master

Issue #4129, opened 2024-03-09
Timeline - Mute is now.. inversed to be "Play Sound"


Draise14:
# Blender
![blender_jp1rx2zO0I](https://github.com/Bforartists/Bforartists/assets/35357409/d09d3dd0-0e97-4a80-9340-c44f7822c20d)

# Bforartists
![bforartists_2HbOr3WL1r](https://github.com/Bforartists/Bforartists/assets/35357409/0a433447-cc14-45c7-bde5-09fe744c7659)

wth, why this change I don't know... but it screwed me over an hour trying to make the sound work, lol

# Task:
1. Rename label to be "Play Audio"
2. Make sure default installation has it "turned on"
3. Keep the API and UX the same as Blender. 
5. Update documentation

Draise14:
Iyad brought up a good point. This toggle is now "Play Audio" in Blender. So if we load a blend file with it's setting stored as activated, it should also activate in Bforaritsts, or unmuted (inversed) in the same file setting.

ReinerBforartists:
They changed it, so we should change it too for consistency reasons. Even if this introduces the same trouble than in Blender. We promised to deliver the same functionality ...

iyadahmed:
> They changed it, so we should change it too for consistency reasons. Even if this introduces the same trouble than in Blender. We promised to deliver the same functionality ...

I think they didn't change it, it was always like this?

Draise14:
Blender changed it recently. 

Yeah, this case would be KISS, and rename the label and make the default play instead of mute. 

I still don't know why they did this....

Draise14:
Sorry bout the confusion @Blenux and burning your time, I should have been more explicit with the task.

Blenux:
All good mate.

Draise14:
Documenting

Draise14:
Commit to master

PR #4157, opened 2024-03-21
Varios commits

Draise14:
#4156
#4153
#4129


Draise14:
Forgot to switch to master, so did it in a branch. Pulling in now.

PR #4243, opened 2024-05-08
fix for 4240-warning-property-mode-not-found-in-item-operatorproperties

Blenux:
-- removed uv sculpt keymap entries causing errors.


esnosy:
Doesn't this change break UV sculpting?
maybe we should see what change in Blender did caused this first,
then if mode is really removed, then we just remove mode, but even then it is not right :/

Draise14:
I will have to check on Monday. They changed these  brushes to no longer use the brush data system I think,to make room for brush assets, an exception to the rule.

Might good to keep and eye on how they did it just in case brush assets UX isn't so great.

Blenux:
So should just remove the mode for now?

PR #4284, opened 2024-05-17
fix for 4279-extensions-add-a-visual-indicator-to-what-is-an-addon-an…

Blenux:
- Added icons to indicate which is what.

image


Draise14:
Was easier than expected, nice. I love how well you know your way and the comments. Great great work.

PR #4303, opened 2024-05-24
fix for 4302-rna_nodetreecc-icon-enumeration-make-order-of-enumeratio…

Blenux:
-- match order of enumeration, the same as in NOD_static_types.h

### DIFF
--- a/source/blender/makesrna/intern/rna_nodetree.cc
+++ b/source/blender/makesrna/intern/rna_nodetree.cc
@@ -11406,6 +11406,8 @@ static int node_type_to_icon(int type)
       return ICON_VALUE_TO_STRING;
 
       /* bfa - GeometryNode */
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Hey Blenux, 

I notice that your commit adds trailing whitespace. Trailing whitespace (empty invisible spacebars or tabs at the end of an expression, or in a new line) is not really dangerous. But disturbing. When you make a commit with trailing whitespace then this is simply in the way of reading the actual code. And it blows up the code a bit too in size. So it's something to avoid if possible. The mean thing is, some editors doesn't even display them.

There are addons for Visual Studio and VS Code that removes this trailing whitespace automatically. I cannot remember which one i have installed here, but it does a miracle at the trailing whitespaces by simply saving the file. Just search for trailing whitespace in the addons. There is more than one around.

PR #4305, opened 2024-05-25
fix 2 for 4276-extensions-expose-the-uninstall-button-to-a-top-level-…

Blenux:
-- Moved Uninstall button next to Installed for Legacy add-ons.


Draise14:
Couple of leading spaces with this commit, but looks fine by me.

Blenux:
fixed whitespaces

PR #4395, opened 2024-06-16
fix for 4392-topbar-file-export-fix-ply

Blenux:
-- updated the topbar ply operators to the built in ones.


Draise14:
Wait. May require to the toolbar editor too.

PR #4399, opened 2024-06-17
fix for 4387-sequencer-text-object-style-box-is-permenantly-disabled

Blenux:
-- Removed Shadow and Outline naming from their props operators. 
-- Float Left for all.
-- Some extra code tidy up (Whitespaces, extra line spaces).


ReinerBforartists:
May i remind to always update the manual with these changes too? Else all these changes are lost in the manual. And then the manual becomes more and more outdated. Which makes the users trouble. They find descriptions that not longer exists. And makes us trouble with updating the manual. A small one liner then means to rewrite the whole chapter since it is outdated ...

PR #4400, opened 2024-06-18
Sync blender main - Week 25

Draise14:


Draise14:
Yes, Windows builds fine here now.

Issue #4491, opened 2024-07-10
Geometry Nodes - Geometry - Write - New Geometry Set Name node


Draise14:
![imagen](https://github.com/Bforartists/Bforartists/assets/35357409/0d039373-4a08-402d-aa39-9f360f7ec7e4)

New Set Geometry Name node, new icon, new documentation.

<details>
<summary>Details</summary>
Geometry Nodes: allow naming geometry sets 
This adds a new `name` member to the `GeometrySet` class. This name can be set
with the new `Set Geometry Name` node. Currently, the name is only used in the
spreadsheet when displaying instances.
The main purpose of this name is to help debugging in instance trees. However, in the
future it may also be used when exporting instance trees or when creating separate
objects from them.
Note, the name is not expected to be unique, it is fully in user control.
Naming geometries is necessary to make the spreadsheet more useful for instances,
because currently the user has no information for which geometry is used by each instance.
We also want to use this name to improve the integration with grease pencil where
sometimes layers become instances with the same name.
Pull Request: https://projects.blender.org/blender/blender/pulls/114910

</details>

ReinerBforartists:
Documented

Draise14:
Looks like the pull request has a merge conflict in the Cmakelists.txt file

PR #4494, opened 2024-07-11
Geometry Nodes - Geometry - Write - New Geometry Set Name node #4491

ReinerBforartists:


Draise14:
This one contains a conflict.

ReinerBforartists:
Ah indeed. I should have merged it after adding new icons ...

Issue #4493, opened 2024-07-11
Remove dat files of the icons in release/datafiles


ReinerBforartists:
The dat files are no more. We can remove them.

![removedatfiles](https://github.com/Bforartists/Bforartists/assets/13919702/4f4d65b3-e941-4c42-807f-94e9f5a2d5fb)

ReinerBforartists:
Done

PR #4495, opened 2024-07-11
Remove dat files of the icons in release/datafiles #4493

ReinerBforartists:


Draise14:
Oh, completely forgot to remove these, will save a few mb.

PR #4503, opened 2024-07-17
Sync blender main - Week 29 

Draise14:
Currently Debug is bugging out, though documents referenced are identical to Blender. So I'm not sure what is up. I will do another merge later this week before release just to make sure it's squeaky clean.


Draise14:
Ok, did some reasearch. The debug build is bugging out with the import STL and OBJ geometry nodes (experimental) because the debug build doesn't compile with OBJ... So will pull to master and resume. The other builds are good.

Issue #3562, opened 2023-06-01
Asset Browser - Add toggle "Instance" to header to avoid linking/appending issues with collections


Draise14:
![bforartists_vmzEhkZEuJ](https://github.com/Bforartists/Bforartists/assets/35357409/0f9c0ed2-df73-47bf-96ec-7779abe37329)

# Issue
Right now when you link or append collections, you have the option in the Repeat Last operator to instance or link to scene. When you link to scene, it will place the container contents to the cursor, then when you switch to Instance, it will offset the center of the collection contents by the offset from world center. There is little or no way to fix this manually and it is a pain in the butt when working with collections. 

On top of that, there is no way to know if you are appending or linking in a collection as an instance or not. It's a wild card with frustrating results. 

# Solution
Ideally, this toggle should be an opt in BEFORE you drag in collections, so you know what you are going to get when you drag and drop. Is it an instance? Will it link to scene? 

If you know it's an instance, you won't accidentally drag in "link to scene" with the global offset, then regret changing it to instance after with the broken offset, no-longer able to use collections as instances when doing layout (necessary center of origin needed). If you know you will "link to scene" ei, not an instance, you know that it will come it ready to be overridden at cursor placement. 

## Task
- Find a way to add the Repeat Last operator to the header so that you can optionally link or append collections with an intentional UX to avoid issues repeat last instance conversion in the future.

iyadahmed:
progress
![image](https://github.com/Bforartists/Bforartists/assets/25552173/4913977f-f37c-4674-b3ad-534cdc3d95bd)

Blenux:
Notice some weirdness with undo when toggling the instance,  and some objects of the collection will be offset a bit.


https://github.com/Bforartists/Bforartists/assets/25260650/ad8319a2-803c-4bb9-96de-1d18ab38b290

ReinerBforartists:
Can this be reproduced in Blender?

Draise14:
Yes, this is an issue about Blender and the way it does placement of object data and how you instace it. Instanecs have COW (center of world) account for any offset of the original mesh data.

Draise14:
Building the test

Draise14:
> Notice some weirdness with undo, and using use instance before and after the collection is already in the scene, and some objects of the collection will be offset a bit.
> 2023-06-04.12-29-41.mp4

[...]

PR #4544, opened 2024-08-07
Initial add exposed asset shelves for 3d viewport and node editor

zNightlord:
Add asset shelf for objects, materials, collections for the 3d viewport, node group asset for node editors (Geometry, Shader, Compositor)
For 3D viewport the default drag behavior would be using instances. 
 `asset_drag->drop_collections_as_instances = true;`
 I can't figure out a proper way to really hook up the buttons, props I made under `AssetShelfSettings` and put in `AssetDragController::create_drag_data`
 
 As for node editors for the default assets, editor filtering I think it's better to define asset [tags ](https://docs.blender.org/api/current/bpy.types.AssetMetaData.html#bpy.types.AssetMetaData.tags) replacement for the current `not asset.name.startswith("S_")` and also for user made node group have to set it up like that for have it properly place in the right node editor type.
 ```py
    @classmethod
    def asset_poll(cls, asset):
        return asset.id_type == 'NODETREE' and not asset.name.startswith("S_")
 ```

image
*ignore the drop instance buttons* 

Extra TODOs for this:
- [ ] Make startup file properly adapt to this
- [x] Adjust versioning code part
- [x]  Typos.


Draise14:
This is a big awesome feature set! 

Maybe for the 3D View, the collection center to world origin and as collection should follow the asset browser settings instead of override?

zNightlord:
> This is a big awesome feature set!
> 
> Maybe for the 3D View, the collection center to world origin and as collection should follow the asset browser settings instead of override?

The asset browser is in a different context space data unfortunately, unless you move it to the Bforartists preferences addon or under preferences.
Related #3562

Draise14:
The properties toggles were set in C, then exposed to the header. But yeah.. might need some thought. I will use the new tag system to filter, so a user can use a tag instead of name. It will clean up the prefix names, not so user friendly anyway.

Draise14:
Work is nearly done, will be able to review it early next week.

Draise14:
Ok, going to pull your branch and do a review. If it seems to work and we have a "default" action for the asset shelf, we can later find new operators to drag things in with the user opt-in (link, append, reuse, etc - then collection or world origin for the 3D View) but that can be new tasks.

Draise14:
Compiled. It works and follows the file Browser settings of linking or appending.

I did find a blender bug though when linking collection to scene, not as instance.

zNightlord:
It's the same operator call as the asset browser one. That actually gives me one idea on how to do something with the button problem for later.

> I did find a blender bug though when linking collection to scene, not as instance.
Hmm, I can't replicate that bug.

Draise14:
Ok, blender fixed the bug, but that would come in the next release. They may have also removed a redundet property we built to compensate the broken UX (dragging in a collection not as instance would offset linked data). 

https://projects.blender.org/blender/blender/issues/126568

PR #4546, opened 2024-08-09
fix for 4529-asset-browser-world-icon-in-filter-is-inconsistent

Blenux:
-- changed world icon

image


Draise14:
Huh, mac build failed?

Blenux:
Strange will re-run it and see, was only a icon change...

PR #4553, opened 2024-08-11
Add Disable search on KeyPress option

zNightlord:
Original code from Lateasusual GooEngine

https://github.com/user-attachments/assets/05349f2b-3819-4da4-881c-96f3a128b6ad

image


Draise14:
I guess we can add this prop to the Interface tab of the user preferences.
Does this also work for the node editor add menus? Or only modifier menus?

zNightlord:
> I guess we can add this prop to the Interface tab of the user preferences. Does this also work for the node editor add menus? Or only modifier menus?

Yes, it should. 
I moved it to interface panel... the python ui of it. It is still under experimental struct in source. I guess you can move it and revert back the padding there.

Draise14:
If this is a GUI feature that's complete, we could move it out of experimental, and document it so a user can opts-in without two toggles.

Issue #4096, opened 2024-02-22
Outliner - Add Duplicate and Duplicate Linked to context menu and hotkeys


Draise14:
Blender listend to twitter, Quackers actually, and we talked of wanting to be able to by default duplicate and duplicate linked directly from outliner. Technically it's possible, but not discoverable. 

Blender added it.. as a hotkey only. 

We should add both hotkey and operators to the context menu to make it discoverable. 

* Duplicate (Shift+D)
* Duplicate Linked (Alt+D)

<details>
<summary>Details</summary>

SHA-1: 537d1752899a37155dfaf4577412298630efc6ce

* Outliner: Add shortcut to Add Object and Duplicate

Add shortcuts to the Outliner for consistency and convenience:
* Add Object (Shift+A)
* Duplicate (Shift+D)
* Duplicate Linked (Alt+D)

More info and images in the PR.

Pull Request: https://projects.blender.org/blender/blender/pulls/117922

</details>

Draise14:
Documenting
![imagen](https://github.com/Bforartists/Bforartists/assets/35357409/02e6f34d-5814-4132-b390-fc8e721a6bcc)

PR #4577, opened 2024-08-21
fix for 4540-hotkeys-sequencer-q-to-quick-menu-is-not-working

Blenux:
-- was accidentally removed in #4096


Draise14:
Big oops 😅😅😅

PR #4579, opened 2024-08-22
fix for 4567-addon-default-addons-has-a-resource-warning

Blenux:
- don't cancel, just return none.


Draise14:
Thanks. I'm learning a lot! But a bit like through the trial and error. 😅

PR #4585, opened 2024-08-24
3645 3d view make subdivisions operators update render subdivisions instead of only viewport

Blenux:
-- added toggle in power user addon to match render levels to viewport level in the Subsurface Modifier.


Draise14:
Had a quick look, seems the grease pencil operators registry to the header menus are commented out?

Blenux:
Yeah, they were throwing an error, need to fix in their own task, haven't got around to it yet.

Issue #4569, opened 2024-08-19
Extensions - Add ability to uninstall extension direclty from the Addon tab


Draise14:
![image](https://github.com/user-attachments/assets/1c24f2c7-d7f0-46e3-8947-fe81f8c9bb4e)

One last attempt here:
- Add the ability to uninstall an extension listed directly from the addon tab

PR #4589, opened 2024-08-25
Extensions - Add ability to uninstall extension directly from Addon tab #4569

Quackarooni:
https://github.com/user-attachments/assets/04677224-1b2b-4030-8bea-50c763d836a2

Implementation is a bit messy for my liking, but this is as minimal as could make the changes to get something working.
Unlike the one for legacy addons, the operator for extensions requires a `repo_index` parameter to work correctly.

The closest place this is calculated is two functions up, so most of the code here is simply for piping that data to the function that draws the button on screen. Did it the same way the manifest data is piped which is through a `dict` with repo indices mapped to module names.


Quackarooni:
Closed this one to rewrite the commit message. A bit pedantic, but ehh might as well.

PR #4591, opened 2024-08-25
fix for 4535-properties-editor-grease-pencil-lineart-modifier-has-fun…

Blenux:
-- added a new lineart icon to the header, which was causing the weird spacing on header.

image


Draise14:
image

Looks like the GP lineart icons need some consistency, consolidation and love accross the board, but will make a new task for that for next release.

Issue #4532, opened 2024-07-30
Node Editors - Sidebar - Move the Copy Color up to top level (single drop down menu entry)


Draise14:
![image](https://github.com/user-attachments/assets/c14d89c1-e065-40df-9cd8-ec41835761aa)
Would be nice if this one operator in the drop down of a hidden panel.. be up a level. I use this a lot.

PR #4592, opened 2024-08-25
Node Editor - Move "Copy Color" operator to top-level (#4532)

Quackarooni:
https://github.com/user-attachments/assets/f882b11c-64ee-4665-bb98-095a4e0bd8ab

Moved the "Copy Color" operator out of the context menu. I put it in a `subrow` layout so that it could be joined together with the color property by setting `align=True`. 

Didn't do that on `row` itself so that the old layout & spacing stays intact, and the context menu can just be re-enabled should other operators be added to it in the future.


Draise14:
Good commenting!

Issue #4564, opened 2024-08-13
Icons - Cleanup icon order in UI_icons.cc and cmake file


Draise14:
Right now the new icon set is one big alphabetical mess - and our system for icons is not alphabetical nor is it concurrent with each other anymore. 

Ideally we can make the SVG files, the UI_icons and the rest mostly in the same alphabetical order.

The only exception is sytem icons that Blender has grouped in a separate way. 

Order is irrelevant now, but we need an atleast consistent thought process here... 

# Task
- Organize most icons alphabetically
- Organize footer icons equally (out of alphabetical order
- Group into letter based chunks so it's easier to navigate or jump alphabetical groups

ReinerBforartists:
Notepad++ offers a sort alphabetically option under edit / line operations. Which should speed up the process.

![sorting](https://github.com/user-attachments/assets/c432a7e7-9635-4d44-a8bd-d8d984c2fae3)

Blenux:
I was only thinking that we should match it close to Blenders has possible and have our icons in appropriate places marked has ours (commented with # BFA, or iconname_bfa etc)

Draise14:
We have over a thousand more than Blender, but the ones that line up to Blenders could indeed make life easier when doing a merge, but in that case the alphabetical order would be lost.

ReinerBforartists:
Alphabetical order makes imho most sense. With over 1000 more icons a comparison with the Blender UI_icons.hh file is impossible anyways. I even fear that a automatical merge could damage our code then.

ReinerBforartists:
Interesting result is interesting

![interesting](https://github.com/user-attachments/assets/7f56ebed-79e3-4e36-a83b-c67b718b78b2)

Draise14:
Oh damn, something got nuked? Maybe there is a key order on some icons?

ReinerBforartists:
Detected some possible trouble makers. I am not sure though if this comments doesn't regard the now gone svg file. But it's worth a shot.

![troublemakers](https://github.com/user-attachments/assets/0f4355a5-ce03-4e8c-bc3b-718134672af3)

ReinerBforartists:
Nope, no problem anymore to change the order here. So, let's continue in small chunks ...

ReinerBforartists:
Smaller chunks needed it seems ^^

![smaller](https://github.com/user-attachments/assets/6967a37f-719d-48c7-b901-034a2ba47990)

ReinerBforartists:
[...]

PR #4613, opened 2024-08-30
Icons - Cleanup icon order in UI_icons.cc and cmake file #4564

ReinerBforartists:


Draise14:
I think I'll do a test compile before a pull, quite a big change! But I think you got it. The alphabet system helps a toooone.

And you were right, thesame order as blender, they use DEF_ICON and we use DEF_ICON_COLOR so there would have been merge conflicts in the whole document anyway.

Blenux:
Linux works fine with these changes,

image

ReinerBforartists:
Yeah this was a mean trap. It's the very first icon that needs to remain at the top of the list :)

Issue #4602, opened 2024-08-28
Shader Node Editor - Missing Shader to RGB in add toolshelf in sidebar


Draise14:
![image](https://github.com/user-attachments/assets/930dd4ef-c58d-49c8-b159-a772d750dd3a)
![image](https://github.com/user-attachments/assets/bd65da73-26e1-4b87-a7ee-af6d7054ec10)

Looks like it's not there in the converter category.

PR #4615, opened 2024-08-30
Shader Node Editor - Missing Shader to RGB in add toolshelf in sidebar #4602

ReinerBforartists:


Draise14:
Simple. 

I am not a fan of this Blender trend to make a new additional APIs for the replacement of the legacy API, instead of overwriting legacy and hitting feature parity with the standing and tested API.

PR #4621, opened 2024-09-04
Sync blender main - Week 36

Draise14:


Draise14:
Hell yeah, mac is fine

PR #4650, opened 2024-09-16
fix for 4649-user-preferences-extensions-uninstall-operator-keep-alwa…

Blenux:
-- set extensions uninstall operator always active

image


Draise14:
Nice, yeah, that helps a tonne. I wonder if we should do the same for the check box to activate the extension.

PR #4651, opened 2024-09-16
Sync blender main - Week 38

Draise14:


Draise14:
What's up with Mac?! I can't test locally.

Draise14:
Looks like it got stuck with the sub-modules of the assets.

Issue #4670, opened 2024-09-23
Asset Shelf - Add view pre-set sizes for thumbnails


Draise14:
![bforartists_xPrRK7CIPf](https://github.com/user-attachments/assets/de40bc8c-be35-4d1e-af84-e59f21b123bc)

Part of the asset shelf workflow requires a round trip to the asset browser to change settings, so the UX is limiting or disconnect. We should add the asset browser options to the header of the asset shelves. 

# Task
- Add view pre-set sizes for thumbnails like in the file browser and asset browser. 
![image](https://github.com/user-attachments/assets/d351617f-99a1-4ac2-8990-1db4aeb03246)

Quackarooni:
Did a bit of looking into this, there's a bit of a hangup, since it seems like `AssetShelf` doesn't have a preset thumbnail size property like `FileSelectParams` does, it only has the integer property. For context, `FileSelectParams` is what is used by both the Asset & File Browser.

| Asset Shelf | Asset & File Browser |
|---|---|
|![image](https://github.com/user-attachments/assets/aa33dbbd-074a-4dbe-827e-5a5b2b16cad5)| ![image](https://github.com/user-attachments/assets/d7df8a6b-a5cb-4793-879d-fd47eaaa1a3a)|

Quackarooni:
Was actually just gonna ask if I could do that, lol. 

Since altering the `AssetShelf` struct in the C++ code itself, is way outside my wheelhouse. But handling that stuff with buttons that just automatically update the integer value under the hood is something I think I can handle.

Draise14:
Yeah, UX is the end game, no worries bout the backend. Python with existing API is easier to maintain anyway.

Draise14:
Will be documented with the new asset shelf chapter I think.

PR #4689, opened 2024-09-27
Asset Shelf - Add view pre-set sizes for thumbnails #4670

Quackarooni:
https://github.com/user-attachments/assets/1ba10c15-808e-4f93-af66-a5168f8f7d15

Added buttons for setting the preview thumbnails of asset shelf to preset sizes.
This is implemented via `ASSETSHELF_OT_change_thumbnail_size` which is a helper operator that edits the current asset shelf's `preview_size` property when executed.

This implementation allows for multiple UI Areas showing the same asset shelf to have different sizes. Other methods I tested such as using an `EnumProperty` stored in `bpy.types.Screen` *(which is the most granular ID type I could use)*, couldn't quite achieve this behavior.

## A Small Compromise
How the buttons are drawn in the UI are different from how the ones in the Asset & File Browser are drawn.
This is a consequence of how the one for asset shelves had to be implemented via operators instead of a built-in enum property like the ones.

| Asset Shelf | Asset Browser | File Browser |
|---|---|---|
 |image | image | image |

Ultimately, I don't think this inconsistency is too bad, and isn't bad enough to warrant not having this feature.
And I think the File Browser is a context different enough that it being different isn't too bad. 

So it's only the difference between asset shelves and the Asset Browser that is mildly jarring.
One option we might have in the future is to use the same implementation used here on the Asset Browser.

### DIFF
--- a/scripts/startup/bl_ui/asset_shelf.py
+++ b/scripts/startup/bl_ui/asset_shelf.py
@@ -4,8 +4,46 @@
 
 import bpy
 from bpy.types import (
+    Operator,
     Panel,
 )
+
+from bpy.props import EnumProperty
+
+
+# BFA - Preset sizes for asset shelf thumbnails
+thumbnail_sizes = {
+    "TINY" : 48,
+    "SMALL" : 96,
+    "MEDIUM" : 128,
+    "LARGE" : 256,
+}
+
+# BFA - Preset sizes and their corresponding UI labels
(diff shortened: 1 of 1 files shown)


Draise14:
I'm ok with this execution. I think the default bforartists theme has these as the same grey to minimize anyway, if I recall. Could be wrong. Good commenting!

Issue #4687, opened 2024-09-26
3D View - Grease Pencil - Edit mode - remove Weight menu from the Grease Pencil header menu


Draise14:
![bforartists_RgQDUjFurz](https://github.com/user-attachments/assets/72da835c-15f2-4439-94fc-18ebd4053bc8)

Weight paint has it's own menu, so this is redundant and useless. Probably a merge accident by yours truly.

PR #4691, opened 2024-09-27
3D View - Grease Pencil - Edit Mode - Remove "Weights" menu from header #4687

Quackarooni:
| Before | After |
|---|---|
| image | image |

Removed "Weights" menu from the Edit Mode header menu of Grease Pencil objects.

The menu is commented out of the code, not erased entirely. 
This is in accordance with other UI items disabled in BFA on the same file.


Draise14:
Yeah, the GP stuff is WIP so not sure what would be "updated" to GP3 yet, or removed completely, so commenting them out. Occassionally will clean it up in the merges.

PR #4693, opened 2024-10-01
fix for 4692-warning-within-comment-wcomment-during-compile

Blenux:
--  fix comment on line 889 that wasn't closed.


ReinerBforartists:
Was just about to comment that the asterisk is at the wrong side of the slash ^^

Blenux:
Yep silly me (recompiled to take a picture of the terminal etc, then spelt it wrong bah!)

PR #4695, opened 2024-10-03
fix for 4680-dopesheet-drop-down-icon-in-channels-is-the-wrong-icon

Blenux:
-- Fix drop down icon in the channels.


Blenux:
Documentation still shows proper icon.

PR #4715, opened 2024-10-06
fix for 4675-properties-editor-grease-pencil-layers-expose-move-to-to…

Blenux:
-- added move up and down operators.

image


Blenux:
Will update documentation when merged.

Issue #4774, opened 2024-10-21
3D View - Grease Pencil - Vertex Paint mode - Split color selector in header


Draise14:
![image](https://github.com/user-attachments/assets/215aa66f-2417-4ccc-8460-234bb66fddfa)

Missing the two color swatche, should be made consistent. 

# Task
- Split color selector in header to show both colors

ReinerBforartists:
Closing since the main task is merged.

Issue #4773, opened 2024-10-21
3D View - Grease Pencil - Vertex Paint mode - Expose color swap to header


Draise14:
![image](https://github.com/user-attachments/assets/215aa66f-2417-4ccc-8460-234bb66fddfa)

The color swatch swap in the brush settings for vertex paint could and should be exposed to the header in a consistent way. 

# Task
- Expose color swap in header

ReinerBforartists:
Closing since the main task is merged.

Issue #4772, opened 2024-10-21
3D View - Grease Pencil - Vertex Paint mode - Update color swatches in sidebar


Draise14:
![image](https://github.com/user-attachments/assets/215aa66f-2417-4ccc-8460-234bb66fddfa)

The color swatch swap in the brush settings for vertex paint is the legacy one still, not the new one - and the two color swatches and color swap could and should be exposed to the header in a consistent way. 

# Task
- Fix color swap in toolshelf to be the new operator

ReinerBforartists:
We have now three tasks for the basically same job. Clean up the color swap mess. It's just not clear which task for what. At least not for me. So who ever claims this task should solve all three at once.

Related: 

3D View - Grease Pencil - Vertex Paint mode - Split color selector in header #4774
3D View - Grease Pencil - Vertex Paint mode - Expose color swap to header #4773

ReinerBforartists:
Closing since the main task is merged.

PR #4784, opened 2024-10-28
Grease Pencil - Vertex Paint Mode - Make color picker UI in header and sidebar consistent

Quackarooni:
https://github.com/user-attachments/assets/840e023c-db21-4ac2-b698-d53219fc3db4

- Added secondary color prop and "Swap Colors" operator to header
- Updated "Swap Colors" operator in sidebar panel to the new version

Opted to not shrink the color swatches the same way as they are in Draw mode, as that requires a lot more layout code and we still have plenty of header space in Vertex Paint mode.


Quackarooni:
Addresses:
- #4772 
- #4773
- #4774

Issue #4779, opened 2024-10-22
Outliner - Colored collection rows


Draise14:
https://github.com/user-attachments/assets/b9674cfa-3681-41d7-8076-a3c2ec40ab31

![bforartists_a2tBfJxDQZ](https://github.com/user-attachments/assets/029fe542-81e7-4eb9-b04b-50db50961851)

For user experience, and out of conversation with users - this seems to be a hot requested feature. And Pablo Vasquez made a prototype. I have since adopted the diff and made an experiment.

I am debating making a toggle for it, and I am now trying to make a design for nested collection to not draw the full width.

Draise14:
https://github.com/user-attachments/assets/aaa858c3-477b-495e-8e9c-bf3f8640b4df


Here is the more final look. Ready for feedback!

ReinerBforartists:
Hmm. Looks decent enough now.

Just curious, isn't this a feature that ist to expect to land in Blender anyways? So why not wait until it does, and then fine tune?

Draise14:
Proposal was vetoed, thus Pablo Vasquez shared it with us directly on Xwitter. Development branch for it got nuked. 

It won't go to Blender.

ReinerBforartists:
Ah, now it makes sense :D

Draise14:
Have tested it all week.. Now I can't go back to the bland outliner without colors. This simply is a win. 

Should I try build in a user prop to toggle colors? Or shall we merge to main? Major 4.3 update.

ReinerBforartists:
Why not both ? ^^

Draise14:
Yeah, seems making a property to implement in C then implement in python is not a simple thing - since this is C change, though aggregate so uninvasive.

Draise14:
I'll merge, then make a new task to make the toggle.

ReinerBforartists:
Yes, that's nothing trivial, unfortunately.

Blenux:
Yeah I like and a toggle is always good for people that don't.

Draise14:
I'll commit this and merge to master, then make a new task for the user opt-out. I am pressed tight for time and will travel Saturday 2nd to the Blendiberia conference, then I present Bforartists there in Spanish, will be back to work on the 14th.

Blenux:
If you make the opt out task, I could do that one.

Issue #4771, opened 2024-10-21
3D View - Brush Assets - Context menu is tied to specials, so needs to be decoupled


Draise14:
Right now the Brush Asset Shelf context menu is.. missing a lot of features making the brush assets context menu really broken. But that was done on purpose to pull things out of the "specials" menu. But the context menu calls the specials menu... hurr durr. 

![image](https://github.com/user-attachments/assets/8069933e-4d68-4a1f-8a7e-c20fb228fe3b)

Let's decouple them.

Draise14:
![image](https://github.com/user-attachments/assets/42135e61-76c7-472c-8664-1ae846b707a1)
![image](https://github.com/user-attachments/assets/d6e14cf5-2ce9-422b-9a78-da63b6d04a27)

Better.

PR #4787, opened 2024-10-30
 Outliner - Colored collection rows #4779 

Draise14:
Added new outliner colors to collections for easier visual "blocking" of complex outliner trees. 

Contains a surface level block, and sub-level L brackets.


Draise14:
Ah, has a fix for another task pushed in? #4771

Did it to the wrong branch, lol

PR #4790, opened 2024-11-02
fix for 4789-show-toolshelf-tab-property-c-code-change-null-to-nullptr

Blenux:
-- replaced NULL (C) for nullptr (C++)


devesnosy (approved):
LGTM

PR #4830, opened 2024-11-30
fix for 4829-important-hotkeys-addon-dysfunctional

Blenux:
- must call the parent function, introduced in [1dbe94c8ac](https://projects.blender.org/blender/blender/commit/1dbe94c8ac) commit

Before

`def __init__(self,)`

After

```
def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```


Blenux:
bah must find why whitespaces get added at random places.

ReinerBforartists:
> bah must find why whitespaces get added at random places.

Most probably by editing the file directly in Blender / Bforartists. I have an addon installed in both, vs and vs code that removes trailing whitespace. And made it a habit to regularly save my files with one of it. Even when i edit the file in Blender / Bforartists.

Issue #4848, opened 2024-12-13
Update Windows Github actions script to be relative


happymimimix:
Windows used back slash '\\' for directory separation! Not forward slash '/'. 
Your build script used the forward slash '/' which is unacceptable in Microsoft Windows! 
The script mysteriously still works however, I have no idea how it still managed to upload a working windows build while we can clearly see windows non-stop complaining about the use of '/' in the build logs. 
![image](https://github.com/user-attachments/assets/c7edde15-cf7f-4bd6-835b-dac602597198)

Also, your build script used absolute paths like C:\Users\runneradmin, which is also a very bad practice! How can you be 100% sure that the username of the runner's machine is runneradmin? You should use relative paths and environment variables to make sure the script works on every machine! 
%CD% gives you the current directory. 
%systemdrive% gives you the drive letter that the system currently runs on. 
The drive letter is not always C:, it could also be Z:, X:, A:, D:, or any other letter! 
%username% gives you the name of the current user. 
%userprofile% give you the directory of the user profile
In addition, not every windows computer has a D: drive! 
A windows computer could have only one C: drive and nothing else, and it would still work just fine. 
You should always check whether the D: drive exists or not before using it in the build. 

To avoid any potential issues with windows builds in the future, I strongly recommend you correct these mistakes in your build script.

ReinerBforartists:
Hey happymimimix,

Many thanks for the report and the hint. We will have a look at it and fix it asap :)
You could make a pull request if you want and have the time for :)
Kind regards
Reiner

happymimimix:
> Hey happymimimix,
> 
> Many thanks for the report and the hint. We will have a look at it and fix it asap :) You could make a pull request if you want and have the time for :) Kind regards Reiner

Let me take a closer look at why this seemingly incorrect script CAN build in the first place. 
This should've gone into a compilation error, at least that's what I was expecting. 
But no, it actually made a working windows build!

happymimimix:
> Hey happymimimix,
> 
> Many thanks for the report and the hint. We will have a look at it and fix it asap :) You could make a pull request if you want and have the time for :) Kind regards Reiner

According to what I found in the logs, this is what actually happened during the build: 

First, the command ```mkdir C:/Users/runneradmin/bforartistsrepo``` is executed. 
Since you incorrectly used forward slash '/' in the command, the command actually does nothing and throws the ```The syntax of the command is incorrect.``` error. 

[...]

PR #4872, opened 2025-01-06
Update Windows Github actions script to be relative #4848

Draise14:
First attempt, not sure how to test.


Draise14:
No dice, the upload can't take a relative.

ReinerBforartists:
Then use absolute paths. Rule number one: it must work ^^

The more vital fix is to remove the part that errors out.

Draise14:
Currently being fixed in another pull request

PR #4875, opened 2025-01-08
fix for 4844-topbar-toolbar-primitive-images-are-greyed-out-and-dont-…

Blenux:
-- removed since drag and drop is supported.


Draise14:
Quick question, was this also removed from the editor topbar?

Blenux:
ah nope forgot about that one

Blenux:
fixed.

PR #4896, opened 2025-01-17
fix for 4855-3d-view---panel-right-click-menu---expose-the-reset-zoom…

Blenux:
-- added the ability to reset the panels via the right click context menu when zoomed.
-- added new icon for it.

(Panel header)

(Buttons in panels)


Draise14:
I think the label may need to be more explicit.

"Reset Panels Zoom" or 
"Reset Interface Zoom"

Or simply:
"Reset Zoom"

"View" is related to the 3D Viewport, or camera view.

Blenux:
oh yep easy fix.

Blenux:
did Reset Panel Zoom, looked the best between all those.

PR #4940, opened 2025-03-10
fix for 4935-3d-view---workbench-overlays---float-depth-of-field-left

Blenux:
- floated Depth of Field to the left like before

image


Blenux:
This one turned into making the specular highlight float left instead of the depth of field (which changed with the recent master merge), keep or close and will redo a specual highlight issue instead?

Draise14:
Make a new issue for the specular highlights if it got nuked

PR #4942, opened 2025-03-10
fix for 4934-missing-icon-view_camera_unselected

Blenux:
-- renamed OUTLINER_DATA_CAMERA to VIEW_CAMERA_UNSELECTED

since addon's that are built for Blender uses this icon instead, resulting in a error when used in Bforartists.


Blenux:
Made a little mistake on this one, will fix soon

Blenux:
should be fine now, at first I overdid it, just needed to add said icon (blender has the icon with that name, we didnt, hence why addon errors out)

(brain is in foggy mode bah)

PR #4947, opened 2025-03-23
4888 sequencer   preview   duplicate operator now possible from the preview mode

ReinerBforartists:
The hotkeys are worth further investigation. I added them manually. It is not possible to add them from the menu. I will check Blender now, and make a bug report. I expect that Blender is not different.

EDIT, forget about it. Blender has the shortcuts also manually added. When you remove it then you cannot readd it back. Which is of course a bug. But i have reported such issues one time too often.


ReinerBforartists:
Documenting.

PR #4991, opened 2025-04-30
Sync blender main znight - Merge Week 18 - take 2

Draise14:


Draise14:
Linux has failed, but I think the file is the same in Blender, so will merge to main, then report as a bug.

Issue #4993, opened 2025-05-01
Linux - Bforartists Thumbnail is crashing builds on Linux


Draise14:
![Image](https://github.com/user-attachments/assets/836ebee4-92cb-408c-87ca-71923f88622c)

![Image](https://github.com/user-attachments/assets/17ce0645-307a-4e3e-afdd-cbfce9a82961)

Seem the Bforartists thumbnailer is causing compile issues on linux, and these files are not directly being compared since the name has changed. Might be a merge issue I did, and was resolved by renaming the thumbnailer, but maybe we should go back to the blender file name then change the relevant Bforartists names again.

Draise14:
Not sure how to test this, my guess was probably wrong, not even the write file that errored. The file is identical to Blender, so a blender build needs to be tested.

Blenux:
did a commit on the pr that will fix this.

PR #4994, opened 2025-05-01
Linux - Bforartists Thumbnail is crashing builds on Linux

Draise14:
#4993


Blenux:
pushed a commit tha should fix this

Issue #4867, opened 2025-01-03
Preferences - File Pather - Re-order library with buttons


Draise14:
![image](https://github.com/user-attachments/assets/329eb95f-32d8-423c-b307-2309f6ebed5e)

Move a library up or down would be... super nice.

Draise14:
Documenting.

Draise14:
Props were missing, so added them.

Draise14:
Documented.

Quackarooni:
> Props were missing, so added them.

Apologies, I did add those when I tested things, but I didn't seem to have added those in the commit.
My bad. 😅

PR #5009, opened 2025-05-01
Preferences - Add operator for reordering asset library paths

Quackarooni:
Implements the feature requested in #4867

How it's currently implemented is a bit of a hack, since the struct containing the asset library paths does not have the functionality for moving items, and thus items can't be re-ordered.

https://github.com/user-attachments/assets/99f10cb9-a963-4f7c-9dee-6f9133375c36

Adding this ourselves in the C++ code would be hard to maintain as we diverge for base Blender. Instead, I opted to make an operator that technically doesn't re-order items, but rather just swaps the data between two adjacent items.

From a user-facing perspective, this achieves the same effect as re-ordering items.


Blenux:
#4994 PR should also fix the failed linux check (thumbnail error)

Issue #5007, opened 2025-05-01
Default Library - Move Grease Pencil node group to own category


Draise14:
![Image](https://github.com/user-attachments/assets/dae2a3bf-eaef-499b-b54f-bfef6447112e)

They added a new grease pencil add category, so the grease pencil node groups of the default library don't need to be in the Curve menu, which makes sense.

Draise14:
There is a blender bug not combining categories right now.

Quackarooni:
> There is a blender bug not combining categories right now.

Yeah, mentioned that in the PR yesterday, the bug was also happening in Blender 4.5.
This task could prolly wait until that gets sorted out on their end. _(hopefully)_

Quackarooni:
Looks like the changes in 00a863d309113cb299e9c5bcc8134a31d5f9f4c0 already applied the same change in categories.
Closing the pull request and this issue.

PR #5012, opened 2025-05-06
Default Library - Move Grease Pencil node group to own category

Quackarooni:
Addresses #5007

Grease pencil assets are moved into the `Grease Pencil` category. 
They are then placed into corresponding subcategories (`Read, Write, or Operations`)

| Read | Write | Operations |
| --- | --- | --- |
| image | image | image |


Quackarooni:
Just a heads up, it seems like there's duplication with nodegroup assets.
It separately draws the top menu, and sub menus, and the sub-sub menus, etc. so menus repeat.

image

Looks like a Blender issue though as it also happened there.

Draise14:
This doesn't make sense that it doubles up. Maybe there is a typo with a space or something.

Quackarooni:
> This doesn't make sense that it doubles up. Maybe there is a typo with a space or something.

I checked, it's not a thing with spaces, it really just duplicates it.
The issue also exists with Blender, possibly a regression.

Quackarooni:
@Draise14 
Also, just noticed that they added a `Set Grease Pencil Color` node natively, and it also controls opacity.
Should our assets for those be deleted? Or do we still keep it?

image

Draise14:
We have stroke or fill filters, so might still be relevant. Maybe we can tag it as "Advanced" with a bit more control and update. I like the drag and drop UX.

New
1. Open asset workspace
2. Make gn tree
3. Add node to tree
4. Adjust

You need 2 or 3 editors to use it... And have to go to the node graph to use it. 

Our method
1. Use asset shelf to drag and drop
2. Viola, adjust

But the nodes could be adjusted to use the new maybe more performance system.

Draise14:
There is a merge conflict with my files though, so I'll update locally then ovrewrite.

Draise14:
Seems like a major regression from blender, so holding off this task for now.

PR #5035, opened 2025-05-13
Sync blender main - Week 19

Draise14:
Note: Many of the modifiers formatting might be botched, but can be fixed on a case by case


Draise14:
Sigh, Mac had a cmake hiccup. My bad..

PR #5080, opened 2025-06-03
Nodes: Add User Preference for Viewer Node Repositioning

Quackarooni:
When the user views an output socket (normally Ctrl Shift + Left Click), 
the Viewer node automatically repositions itself near the clicked output.

Some heuristics have been added in 4.5 to make the auto-positioning less jumpy,
but it's still inconsistent and some users might prefer it to be disable it entirely.

This commit adds a User Preference for toggling this behavior.
It's set to be True by default, matching Blender's normal behavior 
but allowing the user the choice to opt-out if they wish.

image

https://github.com/user-attachments/assets/133ae8db-1738-4499-b816-e8c9da7a60c1


Quackarooni:
This started as a PR on Blender main, but didn't really go anywhere:
[Nodes: Add user preference to toggle automatic viewer positioning #135368](https://projects.blender.org/blender/blender/pulls/135368)

@Draise14 has shown a bit of interest on it, so figured might be worth porting to BFA instead. :D

Issue #5037, opened 2025-05-13
Properties Editor - Render Tab - Grease Pencil - Clarify the SMAA cryptic labels and tooltips


Draise14:
![Image](https://github.com/user-attachments/assets/18bdebfe-f501-4bdd-a5ad-46263a4e71ea)

The GP settings are for artists, and they use an acronym and don't explain it in the tooltip. Make them more explicit and user friendly.. 

It's this... but should be artist friendly:
SMAA (Subpixel Morphological Anti-Aliasing)

Quackarooni:
Found a clearer explanation for SMAA threshold inside the compositor files:

![Image](https://github.com/user-attachments/assets/8c7adf10-d7d5-45cc-a46f-9d3bf2f2a46d)

Will just rephrase this for brevity. And also tweak the other labels and descriptions to be simpler.

Quackarooni:
Documented: 

![Image](https://github.com/user-attachments/assets/7e3f2896-bb12-49db-911d-dd65374c1980)

PR #5081, opened 2025-06-03
Properties Editor - Render Tab - Grease Pencil - Clarify the SMAA cryptic labels and tooltips #5037

Quackarooni:
- Added an `"Anti-Aliasing"` label on top of the AA properties, as it was too long to put in the property labels themselves, but too important in explaining the nature of the props that IMO it shouldn't be left out.
- Renamed `"SMAA Threshold"` to `"Edge Threshold"` & `"SSAA Samples"` to `"Supersampling"`
- Rephrased the tooltip descriptions.

## Panels
image

## Tooltips
|||
| --- | --- |
| SMAA Threshold (Viewport) | image |
| SMAA Threshold (Render) | image |
| SSAA Samples | image |

### DIFF
--- a/scripts/startup/bl_ui/properties_render.py
+++ b/scripts/startup/bl_ui/properties_render.py
@@ -1328,7 +1328,9 @@ class RENDER_PT_grease_pencil_viewport(RenderButtonsPanel, Panel):
         props = scene.grease_pencil_settings
 
         col = layout.column()
-        col.prop(props, "antialias_threshold", text="SMAA Threshold")
+        # BFA - Make clearer labels & descriptions for AA properties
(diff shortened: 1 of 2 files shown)


Draise14:
Maybe the label could go into the panel header instead, so it would say "Render/Viewport Anti-Aliasing" Then the extra label wouldn't be needed in the row. 

I doubt the sub-properties will be outside of AA in the near future.

Quackarooni:
> Maybe the label could go into the panel header instead, so it would say "Render/Viewport Anti-Aliasing" Then the extra label wouldn't be needed in the row.

Got that changed now!
image

Issue #5006, opened 2025-05-01
Compositor - Update node Add tab with Utility and Vector nodes


Draise14:
![Image](https://github.com/user-attachments/assets/a019e266-28dc-4cdc-9345-f817b239f11d)

![Image](https://github.com/user-attachments/assets/f9987762-8a1e-42b3-a7f9-123fee42ef6d)

Update the Utilities and Vector panels from the add shelf. These get new nodes - and document them! If too large, split this task into two.

Draise14:
Adding icons now.

Draise14:
Documenting... this is a big rewrite.

PR #5089, opened 2025-06-06
Compositor - Update node Add tab with Utility and Vector nodes #5006

Quackarooni:
| | Text Buttons | Icon Button |
| --- | --- | --- |
| Utility | image | image |
| Vector | image | image |


Quackarooni:
The new Relative to Pixel node still doesn't have an icon yet, just made this to get the main parts laid out.
Will make a follow-up commit to this branch once an icon is made, so this PR shouldn't be committed until that's done.

Draise14:
Yeah, also the first panel Utility should flow to a grid instead of column, so that would also need a fix there.

Quackarooni:
> Yeah, also the first panel Utility should flow to a grid instead of column, so that would also need a fix there.

That's due to the last button not having an icon, which makes it expand wide and force the flow to be one column.
Should be fixed once an icon is specified for it.

Draise14:
Guess we should do that.

PR #5140, opened 2025-06-29
Fix: 5036-window---expose-window-operators

Blenux:
Added these operators,

Toggle Full Screen
Close Window (along with icon)


Blenux:
Will update manual with these changes when its merged.

Issue #5109, opened 2025-06-20
Geometry Nodes - New Format String node


Draise14:
![Image](https://github.com/user-attachments/assets/dccf0c76-6d2d-477b-9a53-8c327e34c47d)

New node, new icon, new toolshelf panel entry and documentation.

PR #5144, opened 2025-06-29
Geometry Nodes - new Format String node

Quackarooni:
Related issue: #5109

image
image


Draise14:
Will check documentation now.

PR #5146, opened 2025-06-29
Fix: 5107-spreadsheet---move-built-in-attributes-property-to-header-a…

Blenux:
-- removed double template header
-- added show internal attributes toggle to header

used info icon for now/temp if needs changing.

image

will update manual when merged


Draise14:
Reviewing, we have many attribute icons, so will use the default node_attribute one.

Draise14:
image

PR #5172, opened 2025-07-03
Node Editor - Add Menu & Toolshelf - Distinguish nodegroup types with icons and grouping

Quackarooni:
Currently, a nodegroup can either be in these various states:
- Normal : No special behaviors.
- Fake User : Will not be deleted even if it has no users.
- Asset : Marked as asset to be that can be detected by the asset browser.
- Linked : Linked from an external blend file.
- Library Override : Similar to linked, but with overrides specific to the current file.

image

This patch groups each type of nodegroup together and assigns them their corresponding icons. These icons should match the icon being used in the node's group selector. (The one beside the group name).

The change is also applied to the toolshelf panel:
image

### DIFF
--- a/scripts/startup/bl_ui/node_add_menu.py
+++ b/scripts/startup/bl_ui/node_add_menu.py
@@ -47,6 +47,66 @@ def add_node_type_with_outputs(context, layout, node_type, subnames, *, label=No
     return props
 
 
+icon_sorting_order = {
+    'NODETREE' : 0,
+    'FAKE_USER_ON' : 1,
+    'ASSET_MANAGER' : 2,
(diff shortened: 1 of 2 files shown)


Quackarooni:
Note: There is a bit of a bug where nodegroups with library overrides are counted twice. 
(Once as linked assets, and once as library overrides.)

This behavior is also present in main Blender, so I'm just gonna report it there.

Draise14:
Hoooly, this is great!! Good idea.

Draise14:
For rebasing sake in future issues, add some BFA comments, then this is golden to be pulled and documented.

Issue #5160, opened 2025-07-02
Sequencer - Sidebar - Cache Settings - Remove double "Final" property


Draise14:
![Image](https://github.com/user-attachments/assets/a5f41d57-f7e7-4334-a32e-0d2107ec9fec)

Nuff said

PR #5176, opened 2025-07-03
Fix: Sequencer - Sidebar - Cache Settings - Remove double "Final" property #5160

Quackarooni:
image


Draise14:
Thankyou, small but polish is great!

Issue #5170, opened 2025-07-03
Node Editor - Add Menu - Relocate non-add operators in Group submenu


Quackarooni:
Under the `Group` submenu in the Node Editor's Add menu, the following operators exist:
- Make Group
- Insert into Group
- Ungroup
- Toggle Edit Group

![Image](https://github.com/user-attachments/assets/912f85ca-e095-4529-991e-ba60c3c5d8c4)

These operators don't have anything to do with adding new nodes, and instead are about modifying existing nodes. These currently take up valuable screen space and are actually duplicated as they also appear in the context menu.

![Image](https://github.com/user-attachments/assets/42afc58c-31b6-4bb9-a3c6-58ec79921b4c)

In my opinion, it makes more sense for these operators to purely be in context menu, and be removed from the Add menu. I can only speak for myself, but when I call the Add menu, I'd often want to reach for the items here:

![Image](https://github.com/user-attachments/assets/b3f36a20-3661-4507-8ab8-40386b6ffc5b)

So the extra operators just add more travel time for the cursor to reach to the parts I want.

Draise14:
By design we expose all operators to headers for discoverability since context menus are hidden menus, same with hotkey only operators where possible. They exist in the sidebar and header for consistency and discoverability sake here also. (In the relationship tab)

Ah, another note: context menus, sidebar... Are permissible dulplicates but must never be exclusive, as they are opt in or out and hidden by default. 

So will...have to debate the proposal. A compromise is change order of those entries to the end of menu. 

Another compromise is make the relationship tab a higher level in own header entry instead of add menu.

ReinerBforartists:
We have to live with some doubles by design. Since we are dependant of Blender. And can't remove some of them. Addons will go crazy then. 

The context menu is one of these things that i personally have removed at one point as one big double menu. We added it back to be compatible with Blender addons and Blender users in general then. The rmb menu is not easily discoverable. It is against the self explaining and target leading UI UX design that we follow.

Quackarooni:
@ReinerBforartists @Draise14 
To clarify, I'm talking about removing this in the Add Menu. The one you call with `Shift + A`.
IMO they don't fit in with everything else there, as all other operations in the Shift+A menu have something to do with adding a new thing in the nodetree. Either a new node, zone, asset, group, etc.

These operations don't do that, instead they modify existing pieces of the node tree, so they're more apt under the `Node` menu where other operations of that nature go, which is what Blender currently does:
[...]

PR #5182, opened 2025-07-04
Node Editor - Add Menu - Relocate non-add operators in Group submenu #…

Draise14:
…5170

Now exposed to the toolshelf and node header menu in their own sub-group in a consistent way.
Moved the add empty node group to the group panel in the add shelf.


Draise14:
#5170

PR #5191, opened 2025-07-08
Node Editor - Group Sockets -- Improve Add Item dialog

Quackarooni:
Changes made:
- Use `ADD` icon for the popover button
- Create a dedicated "Add Panel Toggle" operator
    - Blender's implementation of having it be a variant of the general "Add Item" operator means that it shares the same poll function as other variants. So for conditions where the other variants are valid, but the panel toggle isn't, the user has to run the operator, have it check whether panel toggles are valid, then it cancels execution and pops a message indicating why it didn't execute.
    - image image
    - This is an anti-pattern, if the operator couldn't run, the button for it should not be clickable. So it makes sense to separate this into a standalone function with its own poll checking.
    - image
- Added helper functions to simplify a bunch of panel toggle related stuff, this'll come in handy for other operators I'll implement later.

### DIFF
--- a/scripts/startup/bl_operators/node.py
+++ b/scripts/startup/bl_operators/node.py
@@ -504,6 +504,46 @@ class NODE_OT_interface_item_new_output(
         name="Item Type", items=(("OUTPUT", "Output", ""),), default="OUTPUT"
     )
 
+# BFA - Create dedicated operator for adding panel checkbox
+class NODE_OT_interface_item_new_panel_toggle(Operator):
+    '''Add a new panel toggle to the currently selected panel'''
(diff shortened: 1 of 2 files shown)


Quackarooni:
I'm keeping the specials menu and adding the missing operators as a separate task, just so this isn't too much too review all at once. Basically just gonna be going at it piece-by-piece.

Draise14:
Sounds good. I'll go through documentation and start updating things. Times like these I wish common documentation was linked like on the web! This past may be instance multiple times per node editor. 

Web 2.0 then documentation 2.0. 🙂‍↕️

Issue #5190, opened 2025-07-08
Interface - Context Menu - Open Drives Editor missing icon


Draise14:
<img width="342" height="394" alt="Image" src="https://github.com/user-attachments/assets/268db77a-3cc2-475b-b7d6-8b7514bafbb0" />

This is a regression, we used to have an icon for the open Drivers Editor operator in the general context menu over properties.

PR #5192, opened 2025-07-08
Interface - Context Menu - Open Drivers Editor missing icon #5190

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

Another chunk of the code was using this as the icon for `"Open Drivers Editor"` so I just copied that.


Draise14:
Yep that was the old icon from documentation

Issue #5189, opened 2025-07-08
Interface - Context Menu - Online Python Refernece needs an icon


Draise14:
<img width="223" height="211" alt="Image" src="https://github.com/user-attachments/assets/8f2b5659-0085-4e38-b426-9279c3d21f99" />

And probably documentation. This is a developer extra. We can use an existing one we use for links or online access.

PR #5193, opened 2025-07-08
Interface - Context Menu - Online Python Reference needs an icon #5189

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

Used the same icon that `"Online Manual"` uses.


Draise14:
Was the icon used in documentation. Nice.

PR #5196, opened 2025-07-08
User Interface - uiDefAutoButsRNA -- Draw Boolean properties aligned left by default

Quackarooni:
Updated the base function that the Redo Panel uses.

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/source/blender/editors/interface/interface_utils.cc
+++ b/source/blender/editors/interface/interface_utils.cc
@@ -403,6 +403,7 @@ eAutoPropButsReturn uiDefAutoButsRNA(uiLayout *layout,
           BLI_assert(label_align == UI_BUT_LABEL_ALIGN_SPLIT_COLUMN);
           col = &layout->column(true);
           /* Let uiLayout::prop() create the split layout. */
+          col->use_property_split_set(!is_boolean); /* BFA - align left if boolean prop */
           col->use_property_decorate_set(!is_boolean); /* bfa - align left if boolean prop */
         }
 


Quackarooni:
This seems to affect a lot of things other than the Redo Panel, but I don't know how to invoke those in BFA.
image
Regardless I think we'd want to align booleans left in most cases, so I don't see the issue in modifying the base function.

Draise14:
Holy you are making many happy here! Had no idea this could be modified globally.. This will have a toooone of work..

Draise14:
How would you modify the base function?

Draise14:
Ah, also adding the split to false on bools?

Quackarooni:
> How would you modify the base function?

The changes on the commit are already on the base function, just felt the need to clarify that. Since looking at the code, it seems like it'll affect more than the redo_panel. But I don't think that's bad as we'll prolly want to left-align in as many places as we could.

What I did was just added the condition to only enable `use_property_split` if the property is not a boolean.
That left-aligns them since `use_property_split` is what does the centering for booleans.

ReinerBforartists:
I don't like this centralized way at all. It makes it super complicated to indent props now. I hope it is not too much in the way ...

Draise14:
image

Seems to be a non-issue.

Draise14:
This is a one-liner, so if we notice things going wild or too hard, we can always revert. I'm gonna greenlight this one.

PR #5197, opened 2025-07-08
Revert "User Interface - uiDefAutoButsRNA -- Draw Boolean properties aligned left by default"

Draise14:
Reverts Bforartists/Bforartists#5196


Draise14:
Oops. Forgot to test where there are indents.

Draise14:
image

It's a non issue.

ReinerBforartists:
This is python. I am more worried about the c props where we did some intendation.

Draise14:
Ah ok, if we see C props where this is required (a label), we can revist. I can't recall where they would be placed. I guess this would be mostly in the Last Operator panel? 

If you see where it breaks, make a new task @ReinerBforartists and I'll revisit , if not, this will most likely save up some main power on maintenance and still keep things mostly floating left on Bools.

PR #5213, opened 2025-07-11
Fix: 5212-toolbar-editor---toolbar-manager-broken

Blenux:
-- change from short to int with needed padding adjustments


ReinerBforartists:
Good catch. Didn't notice.

Issue #5219, opened 2025-07-11
Remove the term Blender from the translation files.


ReinerBforartists:
We never cared about the localization files. It was back in the days too much work to update them with every merge. 

@Blenux brought up the idea to do a simple search n replace to replace at least the term Blender in the translation files. This is the task for it. 

Related: https://github.com/Bforartists/Bforartists/issues/5210

ReinerBforartists:
Simple grep search and replace Blender by Bforartist. Make sure you replace upper case. Lower case would damage the language team url.

I was worried for a short moment since the project id version is then not longer Blender, which i thought could make trouble. But looking good at my end. Tested with german language. But please check with other languages too.

Make sure you replace upper case. Lower case would damage the language team url.

![Image](https://github.com/user-attachments/assets/8006f4db-f0ba-481f-b438-9464e63e295e)

![Image](https://github.com/user-attachments/assets/759f0911-70af-4888-8c7c-93642d54fb81)

PR #5221, opened 2025-07-11
Remove the term Blender from the translation files. #5219

ReinerBforartists:


Draise14:
Looks fine by me. Linux usually takes longer, so will pull.

PR #5223, opened 2025-07-12
Fix: 4853-editors---tabbed-toolshelf---width-is-not-changing-when-on

Blenux:
- This fixes the long standing issue of the Toolshelf tabs not updating.

https://github.com/user-attachments/assets/91520ded-6ee2-4267-a7ec-bd0b8192a42b


Blenux:
small hiccup, will fix!

Blenux:
All good now

PR #5255, opened 2025-07-15
Asset browser import method mode buttons consistent

### LINKED ISSUE #4984: Asset Browser: Make drop to origin button consistent
Image

Image

Also there's no reason to hide it when no collection is selected

zNightlord:
Currently it uses an operator checkbox to set the two stage together. 
This add a set function prop for `instance_collections_on_append` to also set `instance_collections_on_link`

Resolved #4984


Draise14:
Yeah, much more elegant.

PR #5266, opened 2025-07-16
Node Editor - Group Sockets - Move Item operator - Update with 4.5 changes

Quackarooni:
- Fix issues when moving sockets to panels with boolean toggles.
- Support popping panel in/out of other panels.

The code got significantly bigger, as the operators no longer rely on implicit behavior of `interface.move` and is instead handling all cases explicitly in the Python code. This fixes the issues mentioned above, and should also make it easier to account for future cases where they might add new item types.

## Before
https://github.com/user-attachments/assets/fafc2e74-8cef-450e-8c50-079e1914daef
## After
https://github.com/user-attachments/assets/2e477189-5658-452a-aa90-28cce4f861a9

### DIFF
--- a/scripts/startup/bl_operators/node.py
+++ b/scripts/startup/bl_operators/node.py
@@ -58,6 +58,39 @@ def get_panel_toggle(panel):
     except (AttributeError, IndexError):
         return None
 
+
+#BFA - Helper functions for detecting different group item types
+def is_input(item):
+    return compare_attributes(item, item_type="SOCKET", in_out="INPUT", is_panel_toggle=False)
+
+
+def is_output(item):
+    return compare_attributes(item, item_type="SOCKET", in_out="OUTPUT", is_panel_toggle=False)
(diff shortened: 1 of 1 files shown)


Draise14:
Nice.

Draise14:
I think I can pull this one in, but may need a few more comments throughout to make sure this doesn't get lost in a rebase or future merge.

Quackarooni:
> I think I can pull this one in, but may need a few more comments throughout to make sure this doesn't get lost in a rebase or future merge.

The `Move Item` operator already has a `BFA - ...` comment on both its start and end, it doesn't show up in the diff unless you show more cuz I didn't change that. 
image
image

Is that enough already? Or should I also add comments for each of the class methods there?

Draise14:
Ah, didn't seem. More the merrier 🤣 

But Les go. I think tis fine with header footer comment.

PR #5285, opened 2025-07-17
Set property split for importers, exporters

### LINKED ISSUE #4221: Exporters/Importers - Float Properties check boxes Left
bforartists_rhiGmKtcJz

Spent all afternoon trying to do this.. but this is heavily set in C++ and C... 

Goal here is to float the booleans left by making them go to a new row line. If this can be done, in theory, it would reflect for all importer and exporter panels both in the File Browser and the Properties Editor - Collections Tab - Exporters Panel properties. 

Could be here....: 
Cursor_ox4E1KTFCv

zNightlord:
This adds split to the importers/exports, checkboxes left.
Resolved #4221


zNightlord:
image
image

Issue #4997, opened 2025-05-01
UV Editor - Overlays - New Display slider label indentation


Draise14:
![Image](https://github.com/user-attachments/assets/c6996b7a-0275-4676-a6e4-b6c9bbf239f2)

You can show UV's now in object mode, so it has a slider now, and an overlay setting - needs the prop to collapse and the slider label go out, then be documented.

PR #5287, opened 2025-07-17
UV Editor: UV DIsplay in object mode context

zNightlord:
Add collapse button and indent to the UV slider
Resolved: #4997


zNightlord:
image

PR #5306, opened 2025-07-19
Fix: 5102-add---empty-image---remove-from-image-menu

Blenux:
- Removes the operator_menu_enum (which was causing the image to appear in the empty menu)
- Removes the empty image from the empty menu

This matches blender both in menu and codebase.


ReinerBforartists:
Manual is updated?

Blenux:
Yeah can push it when this is merged etc.

ReinerBforartists:
Just do it anyways. That way no change gets lost :)

Issue #5310, opened 2025-07-19
Grease Pencil Edit mode - Stroke menu - Simplify Stroke


ReinerBforartists:
Task: Replace the four operators here with the Simplify Stroke operator from the right click menu. Same functionality, but makes the menu slimmer.

![Image](https://github.com/user-attachments/assets/041e9b8c-619c-42de-9fac-a430b11c965c)

ReinerBforartists:
Blender also has this two different methods.

![Image](https://github.com/user-attachments/assets/8bb45684-27c1-45e6-8f96-c04f47b3b3ad)

ReinerBforartists:
Lemme fix this one too, then i don't have to touch the manual twice.

ReinerBforartists:
Documented

Draise14:
Would rather have these exposed to a top level as 4 separate operators consistently.

Too much digging as a single operator, and not everyone uses the repeat last operator to change these. So we hide functionality from the user, and add an extra step to 4 very different operations. 

Per the Bforartists way, like the UV unwrap methods, we try expose all these to the GUI. Not hide.

ReinerBforartists:
Will revert to have these operators separated. 

But by having it top level we throw the user to death with options, which makes it harder to find what you really need. You loose with searching what you win with putting it top level. And it is not clear that these four operators belongs together. With a sub menu it is. This is the better way imho.

ReinerBforartists:
Okay, doing it the Blender way means to add icons to the enum. It's a C menu then. Easy enough.

![Image](https://github.com/user-attachments/assets/dc69c17e-6b1e-4707-ac6b-bdd4071cd4e4)

ReinerBforartists:
... until you try. Where is this enum -.-

ReinerBforartists:
Hmm, the tooltips are the same for all four operators. Another task.

![Image](https://github.com/user-attachments/assets/4d4fe5b3-6446-431c-9452-d9b705fe13e3)

ReinerBforartists:
Okay, enum is not to catch. let's solve it with a python menu instead.

ReinerBforartists:
Documenting.

@Draise14 , in case you really want to have it as top level as possible, then you could put the single operators into the sidebar. That's still the one click operator at the moment. Not sure if these tools work from the toolbar, but that's another option too. The toolbar shows amazingly few buttons here.

[...]

PR #5311, opened 2025-07-19
Grease Pencil Edit mode - Stroke menu - Simplify Stroke #5310

ReinerBforartists:


Draise14:
Instead of minimizing (and hiding from user), these should be exposed to GUI in a consistent way. So the 4 simplify methods should be top level.

ReinerBforartists:
So the other way around? Okay. I'll change it.

Top level makes the menus too crowded though. And that way it is not clear that these menu items belongs together. We would need to extend the labels. The Blender way is better here. Don't throw the user to death with all the options at once. UX design Rule says not more than seven items at once. And this means the sub menu is better here. Even when it is one more click. You don't do this click every second. And if you dou you can always hotkey the operator.

PR #5315, opened 2025-07-20
Toolshelf - Node Editors - Complete refactor of the toolshelf code

Quackarooni:
Currently, there are numerous instances of entries in the toolshelf either being missing. inconsistent, or misspelled.

A lot of this stems from the fact that there's a lot of manual work + boilerplate associated with adding an entry to the toolshelf.
Especially considering each entry has to be added either 2 or 3, once for each version of the UI. _(Text Buttons, Icon Buttons, or in shader specifically, Common)_

This patch aims to make the process of adding entries to the toolshelf by introducing a unified `draw_entries` function, that handles drawing both text and icon buttons, depending the mode specified by the user. 

An `OperatorEntry` data class is also introduced to automatically handle the various manual previously done, such as specifying the icons, labels, and padding for left-alignment. These are now taken in as parameters by the class, or calculated automatically in cases where they aren't specified.

Handling the visibility of various entries have also been improved, now done via a `poll` parameter in `OperatorEntry` that `draw_entries` skips over if `False`. 

The various `_common` panels have also been removed, as this can be handled by dynamically switching the list of entries passed inside of `draw_entries`. Not having separate panels for these eliminates that jarring "teleporting" effect, where a panel and it's "Common" counterpart are located in different positions in the sidebar, making it seem like the panel is jumping to different place when the `Common` property is toggled in the UI.


Draise14:
Quite a big one to review. Will compile it when back at the studio.

Quackarooni:
@Draise14 Apologies for the massive commit. 😅
Admittedly a lot of it is just getting rid of repetitive boilerplate.

I would recommend not reviewing it through the github diff, as it's basically a complete rewrite, so the diff is super messy.

If you checkout the branch itself, I think that better conveys how much saner the code has gotten. So it'll be easier to make future changes and add entries to this, and we really shouldn't have to do any massive overhaul like this any time soon.

| Before | After |
| --- | --- |
| image | image |

I do kinda see this as a "rip the band-aid off" moment, where it's better to just get rid of these compounding issues at all once, as grueling as that may be. Rather than doing this piece by piece and taking so long to finish.

Draise14:
Yeah! Sounds like easier maintenance in the future. 

I'll compile early tomorrow before the release branch. 

I wonder if this same philosophy could be applied to the tool side bars. They too have 3 or 4 instances of the same operator and icon to draw or each column. 

Nice work, and more sane.

Draise14:
image

Ok, reviewed @Quackarooni . Looks good, no errors, feature complete - only a few observations:

1. The refactor could have a bit more "comments" or instructions as to how to maintain in the document, explain what does what briefly - incase someone noobish pokes around. The pad with arbitrary integers, not sure if that will make of break adding new nodes - should be clarified! How do you define them? How do you add the correct `pad=[integer]` per new node?
2. The group operators are now a regression as they were moved from the sidebar to the toolshelf as they are now node operators instead of add operators. 
3. A comment or two to show that this is a Bforartists or "BFA" somewhere document for keep seeking just in case it gets missed by a future rebase. Tagging a document with those three letters helps see what is being rebased. 

So what I'm going to do now, as this solves a number of QA we need for the release, is I will pull this to master, remove the group panel under Relations in the branch prior, and let you do the rest for the future, as this is not user facing observations. I'll open a new task there. 

Gonna give you a fat bounty for this one! Nice effort!

Draise14:
Dammit, nuked the wrong thing.

Draise14:
Les gooooooo

PR #5320, opened 2025-07-20
WIP: UV Select Similar menu header

### LINKED ISSUE #3204: UV Editor - Edit Mode - Dynamic Select Similar menu needs to be in Header
image

The Select Similar operator in the header of the UV Editor needs to be dynamic to show the operators. 

This is a continuation of #3088 - maybe another hand can make this aspect work.

zNightlord:
UV Select similar in the source is dynamically change based on `uv_select_similar_type_itemf` function in `uvedit_select.cc` which makes hard to exposed options as a menu without populating the enum first. 

The idea of this is add a new Select similar operator with an internal flag which make it can not be search through Operator Search and more importantly without that dynamic change like the original operator and no tampering with the original operator.

https://github.com/user-attachments/assets/44851852-7057-42a3-a542-a3691d153058

Resolved #3204 

*keep this as a WIP for now*

### DIFF
--- a/scripts/startup/bl_ui/space_image.py
+++ b/scripts/startup/bl_ui/space_image.py
@@ -249,6 +249,42 @@ class IMAGE_MT_select(Menu):
         layout.operator("uv.select_more", text="More", icon="SELECTMORE")
         layout.operator("uv.select_less", text="Less", icon="SELECTLESS")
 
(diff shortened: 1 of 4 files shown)


Draise14:
Looking much more usable from the GUI.

PR #5324, opened 2025-07-21
Add draw fill guides info bar indicator 

### LINKED ISSUE #5208: Grease Pencil , Draw mode, Fill tool - make Fill Guide discoverable
When you are in draw mode with a grease pencil object, and when you have the Fill tool activate, then you can draw a Fill guide with holding down Alt and drawing with the left mouse.

This feature is not to discover from the UI. Somebody has to tell you about. And that's bad UX.

No idea how the solution could look like. But we should make it discoverable.

https://docs.blender.org/manual/en/4.5/grease_pencil/modes/draw/tools/fill.html#fill-guides

Related issue: https://github.com/Bforartists/Bforartists/issues/5165

Image

zNightlord:
Add draw fill guide keymap display for the info bar when select fill type brushes

Resolved #5208 

image

### DIFF
--- a/source/blender/editors/interface/templates/interface_template_status.cc
+++ b/source/blender/editors/interface/templates/interface_template_status.cc
@@ -13,6 +13,7 @@
 #include "BKE_global.hh"
 #include "BKE_layer.hh"
 #include "BKE_main.hh"
+#include "BKE_paint.hh" // bfa grease pencil fill
 #include "BKE_report.hh"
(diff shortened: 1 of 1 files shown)


Draise14:
Oh so that's how you do it! Nice!

ReinerBforartists:
This cound be something for upstream by the way.

Draise14:
This is true.

Issue #5335, opened 2025-07-23
Asset Shelf - Thumbnail sizes need consistency with defaults


Draise14:
# Proposal
<img width="1062" height="530" alt="Image" src="https://github.com/user-attachments/assets/627ac9b2-678d-4795-bde1-475a57d5038b" />

It would be nice to have the "Tiny" preset the same size as the default toolshelf size. And the "Medium" size the default asset shelf size when you first open it. Right now the buttons don't "reset" to the medium size so it feels disconnect once you toggle the buttons. 

This would also need to be applied to the asset browser

**Suggested:**
- Tiny: 36 (new: same as toolshelf)
- Small: 64 (new: asset shelf default)
- Medium: (unchanged, asset browser default)
- Large: (maximum, unchanged)

# Problems shown
<img width="921" height="519" alt="Image" src="https://github.com/user-attachments/assets/f93905b1-0f59-4c1c-afc7-a2cff72c306f" />
Small version is biiiig

<img width="975" height="510" alt="Image" src="https://github.com/user-attachments/assets/d89e65aa-d42c-4d91-8503-81a86f87c2a1" />
Medium is also big, normal reset is not there.

Blenux:
I see what you mean, lets me have a looks.

Blenux:
With this

> Tiny: 36 (new: same as toolshelf)

Did you mean 32?

<img width="358" height="46" alt="Image" src="https://github.com/user-attachments/assets/110f4323-32c3-49df-82b3-876f32bdf847" />

That way the sizes would be double for each preset

32 64 128 256
Tiny Small Medium Large

Blenux:
pushed, can adjust if needed

PR #5341, opened 2025-07-25
Fix: 5335-asset-shelf---thumbnail-sizes-need-consistency-with-defaults

Blenux:
-- adjusted default sizes mention in issue #5335 

asset shelf default = small (64)
asset browser default = medium (128)

Would this need to updated in manual?


Draise14:
This is not really a user facing operator or preference, and would only affect screen captures - which would require a refactor... which is underway, so no need.

PR #5371, opened 2025-08-01
Fix: 5365-header-right-click-menu-at-tabs---remove-menu-item-reset-pa…

Blenux:
-- Make the Reset Panel Zoom operator only appear in the correct places

Toolbar
Sidebar
Properties Editor

example - no longer shows in topbar or headers etc
image


Draise14:
This C++ stuff and examples.. it's slowly dawning on me. Nicely done! Thanks for this.

Issue #5322, opened 2025-07-21
Toolshelf - Node Editors - Add comments to the code file for better maintainability


Quackarooni:
Some dev-facing stuff mentioned by @Draise14 in https://github.com/Bforartists/Bforartists/pull/5315#issuecomment-3096892347.

## TODO
- [ ] Add comments clarifying that padding is manual and should be tested & tweaked when adding a new node.
- [ ] Add BFA comment to tag the whole file as custom to Bforartists.

PR #5372, opened 2025-08-01
Toolshelf - Node Editors - Add comments to the code file for better maintainability #5322

Quackarooni:
This really shouldn't make any functional difference, as it's just comments added to the Node Editor toolshelf file.

### DIFF
--- a/scripts/startup/bl_ui/space_node_toolshelf.py
+++ b/scripts/startup/bl_ui/space_node_toolshelf.py
@@ -12,6 +12,8 @@ from nodeitems_builtins import node_tree_group_type
 from .node_add_menu import draw_node_groups, add_empty_group
 
 
+# BFA - Custom panels for the sidebar toolshelf
+
 # Null object used to abstractly represent a separator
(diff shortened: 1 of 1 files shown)


Draise14:
Quick question, is the padding a sequential integer? Does it have to be higher than the highest when adding? Or does it have to be one that hasn't been used?

Quackarooni:
> Quick question, is the padding a sequential integer? Does it have to be higher than the highest when adding? Or does it have to be one that hasn't been used?

Not really, it depends on the length of the node's name.
Whatever is needed to align that node's label to the other items in that panel.
Longer item names need less padding, shorter ones need more.

The whole idea is that `pad` refers to how many spaces gets added to the label text. 
Since the UI font isn't monospace, you can't just count the letters and predict the needed padding. 
Only choice is to basically trial-and-error different padding amounts until the text for the new node is visually left-aligned with the other items.

Draise14:
Guess that little note would be good in the comment in the header or something.

Issue #5389, opened 2025-08-05
Sequencer - Sidebar - Reverse Frames should float left


Draise14:
<img width="539" height="294" alt="Image" src="https://github.com/user-attachments/assets/e674f7f3-0896-4a15-8c55-7e2695aee7e6" />

This is a regression.

PR #5406, opened 2025-08-10
Sequencer - Sidebar - Reverse Frames should float left #5389

Quackarooni:
| Before | After |
| --- | --- |
| image | image |


Draise14:
Documentation already updated, this was a regression.

Issue #5394, opened 2025-08-07
Mesh object in edit mode, Face sub mode, context menu - double separator


ReinerBforartists:
We have a double separator here. Either there is content missing, or one separator line is too much.

![Image](https://github.com/user-attachments/assets/0c42cd4f-6f25-4315-b37b-27a708634772)

PR #5409, opened 2025-08-10
Mesh object in edit mode, Face sub mode, context menu - double separator #5394

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

A separator meant to be added to `col` was added to `layout` instead, meaning it got added at the end of that specific section of the context menu. So this doubles the separators as there is already a separator already intended there.

### DIFF
--- a/scripts/startup/bl_ui/space_view3d.py
+++ b/scripts/startup/bl_ui/space_view3d.py
@@ -7151,8 +7151,6 @@ class VIEW3D_MT_edit_mesh_context_menu(Menu):
             # BFA-Draise - Legacy Operator Added to own group with consistent order
             col.operator("mesh.inset", icon="INSET_FACES")
 
-            col.separator()
-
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
Do you please think about the manual too? :)
Six pull requests means we have minimum five undocumented changes again. And i have just started with the refactor ...

Quackarooni:
> Do you please think about the manual too? :) Six pull requests means we have minimum five undocumented changes again. 

4 of these pull requests are fixing UI issues, i.e. aligning properties left and removing double separators.
These don't have anything to document IMO.

I'll make a note to update the manual about the changes to the `Curve` Menu, and added text styling hotkeys. But I'll have to wait until these changes get merged.

ReinerBforartists:
I have to disagree. The shots in the manual needs an update with these changes.

Draise14:
Yeah, might be easier to track if the documentation is done for the pull request - push that forward.

Issue #5368, opened 2025-08-01
Hotkeys - Text Object - Add common software hotkeys for bold, italic and underline


Draise14:
<img width="391" height="397" alt="Image" src="https://github.com/user-attachments/assets/998d4beb-b71c-4a80-8160-c8e2f47f3465" />

All text editing software have these hotkeys, CTRL+B for bold, etc. We should have them too to follow this standard and ease of use. Mostly everyong has to define their hotkey here out of the box when it's customary to do this with the hotkey, think.. Inkscape, Word, Libreoffice, WPS, Microsoft 365, Wordpad, Markup editors, you name it.

PR #5410, opened 2025-08-10
Hotkeys - Text Object - Add common software hotkeys for bold, italic, and underline #5368

Quackarooni:
image

Added hotkeys for text styling in both the `Bforartists` and `Bforartists-macOS` keymap presets.
The assigned shortcuts are copy-pasted from the Blender keymap and are as follows:
- Bold : `Ctrl B`
- Italics : `Ctrl I`
- Underline : `Ctrl U`
- Small Caps : `Ctrl P`


Quackarooni:
I don't feel strongly as to whether I should've included small caps, or just kept it only bold, italics, and underline.

But I figured that they're classified all text-styling operations and it doesn't make too much sense to leave out just one specifically.
Also, all the other keymaps have it set to `Ctrl P` as well, both Blender and Industry Compatible.

PR #5414, opened 2025-08-12
Draw collapsed node icon

esnosy:


Draise14:
Resolved through work done through this pull request #5425

Thanks for this! Made you a dev here again @esnosy

Issue #5404, opened 2025-08-10
Node Editor - Change socket shape of `Single` data structure to square


Quackarooni:
The socket used for a single value in the Node Editor is currently a thin rectangular socket.
This sticks out a bit since it takes up less visual compared to other shapes.

The proposal is to change this socket shape into a square, balancing it with the other socket shapes and making it either to notice and click.

| Before | After |
| --- | --- |
| <img width="1566" height="821" alt="Image" src="https://github.com/user-attachments/assets/7542d810-95e0-40da-a3e3-c534dc1bd199" /> | <img width="1566" height="821" alt="Image" src="https://github.com/user-attachments/assets/5e850c14-2676-4452-b81c-f35e28412432" /> |

Quackarooni:
This was something @Draise14 requested. 
I personally think it's sensible, the rectangular sockets do look a bit too thin IMO.

It might be worth asking community members though how they'd feel about this change.

ReinerBforartists:
Looks a bit more busy now, but i like the change.

Draise14:
I'll run a poll in discord and x/Facebook Monday. 

But yeah as mentioned, I like square more. Easier to grab and see, more standard.

PR #5429, opened 2025-08-15
Node Editor - Change socket shape of Single data structure to square #5404

Quackarooni:
This patch changes the socket shape for Single Values in Geometry Nodes.

The original one in Blender is a thin rectangle. While for BFA, the choice has been made to draw this as a square.
This gives the sockets more screen space, making it more proportional to other socket shapes.

|  |  | 
| --- | --- |
| Before | image |
| After | image |

### DIFF
--- a/source/blender/gpu/shaders/gpu_shader_2D_node_socket_frag.glsl
+++ b/source/blender/gpu/shaders/gpu_shader_2D_node_socket_frag.glsl
@@ -99,7 +99,7 @@ void main()
     /* BFA - Change single socket shape to enlarged square*/
     /* The name _LINE is still kept though for compatibility */
     case SOCK_DISPLAY_SHAPE_LINE: {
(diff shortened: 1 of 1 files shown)


Quackarooni:
A thing to note, I changed the scaling from 1.2x to 1.1x, in case you notice they've changed a bit.
 
So the square sockets are a tiny bit smaller than my initial screenshots in Discord, but IMO this feels more balanced where it isn't too big compared to other sockets.

Draise14:
Guess this will need to update all geometry nodes and compositor nodes screen captures in the refactor, so will close this for now.

Quackarooni:
> Guess this will need to update all geometry nodes and compositor nodes screen captures in the refactor, so will close this for now.

Note, we'd have to wait till the new socket shapes are moved out of experimental.
As far as normal user with no special settings is concerned right now, the socket shapes haven't been changed yet.

PR #5430, opened 2025-08-16
Fix: 5424-mesh-in-edit-mode---mesh-edit-mode-overlays---remove-double…

Blenux:
This removes the duplicated row of buttons with no icons.

image


Blenux:
Have updated image for manual ready to go.

PR #5432, opened 2025-08-16
Cleanup: Header - Move method for drawing editor type selector from `Menu` to `Header`

Quackarooni:
Menu classes have a helper class method for drawing the editor type selector. This is then later called by headers to draw said selector.

This is not an ideal abstraction, as a menu for each editor type is registered. 
These menus all do the same thing, but each has to registered so that the header for that area can use their class methods. 
This also required littering the menu code with comments explaining what they're for and that they can only be used in headers. Overall, it's just a very roundabout way to implement things.

All of this can be better implemented by moving the helper method to the `Header` class instead, eliminating the need to register menus and simplifying the process for drawing the editor type selector.

### Technical Details:
- Changed `Menu.draw_hidden` to `Header.draw_editor_type_menu`
- Removed the following menus:
    - `ALL_MT_editormenu_clip`
    - `ALL_MT_editormenu_console`
    - `ALL_MT_editormenu_dopesheet`
    - `ALL_MT_editormenu_filebrowser`
    - `ALL_MT_editormenu_graph`
    - `ALL_MT_editormenu_image`
    - `ALL_MT_editormenu_info`
    - `ALL_MT_editormenu_nla`
    - `ALL_MT_editormenu_node`
    - `ALL_MT_editormenu_outliner`
    - `ALL_MT_editormenu_properties`
    - `ALL_MT_editormenu_sequencer`
    - `ALL_MT_editormenu_spreadsheet`
    - `ALL_MT_editormenu_text`
    - `ALL_MT_editormenu_toolbar`
    - `ALL_MT_editormenu_view3d`

### DIFF
--- a/scripts/startup/bl_ui/space_userpref.py
+++ b/scripts/startup/bl_ui/space_userpref.py
@@ -46,7 +46,7 @@ class USERPREF_HT_header(Header):
         layout = self.layout
         layout.operator_context = 'EXEC_AREA'
 
-        layout.template_header()
+        self.draw_editor_type_menu(context)
 
         USERPREF_MT_editor_menus.draw_collapsible(context, layout)
 


Quackarooni:
Just to reassure that this change doesn't break anything, here's a recording of all the different editor's headers still working the same:

https://github.com/user-attachments/assets/2e712283-4464-4c57-addb-ca4ce7496480

Yes, apparently `Preferences` not respecting `Hide Editor Type Menu` is also present in the old code, 
I only learned of this while recording this video. I'll just fix it in this patch as well as the changes here makes doing that really simple.

Blenux:
Very nice., much cleaner code.

PR #5446, opened 2025-08-21
Fix: 5445-properties-editor---render---volumes-panel---float-unbiased…

Blenux:
- Float Unbiased to the left

image


Blenux:
Have updated Manual Image and description ready.

PR #5455, opened 2025-08-23
Cleanup: Fix "alight left" typo in BFA comments

Quackarooni:
Some comments say _"alight left"_ instead of _"align left"_.

This would make them not show up when searching _"align"_.
That is not ideal in the case where we'd wanna track all cases where we change code to align properties differently.


Draise14:
Late Friday night woops

Issue #5355, opened 2025-07-28
Status Bar/Splashscreen/About - Show "Sub Versions" in easier addon and Blender compatibility


Draise14:
<img width="1595" height="614" alt="Image" src="https://github.com/user-attachments/assets/e830966c-3121-41ec-99f8-13897579e8da" />

Right now we have a slightly different way of versioning than blender, but has been a re-occurring theme between addon development and user support. It gets confusing since we are always on the main trunk, but release with versinioning from a branch. I know it's arbritrary numbering - but sometimes sub-versions in a render pipeilne or addon compatibility is useuful and necessary for a smooth file compatibility and addon compatibility, and also Blender compatibility. 

So.. we should show subversions to the user on demand, either in splashscreen or in the footer versioning and about versioning. 

This will also help us "version" development builds when users report issues in discord and such.

Draise14:
Ah, needs to also be added to preferences for consistency.

PR #5464, opened 2025-08-25
WIP: Bforartists version prompt adjustment, status bar Blender version

### LINKED ISSUE #5356: Overwrite Warning - Add explicit versioning information about Bforartists and Blend files (not "Blender")
Image

Right now with file versioning, and when we overwrite with another file version, it mentiones Blender which is asynchronious with the Bforartists versioning and doesn't talk of blend files by Blender versions - so.. we should clarify this a bit and be more on brand.

zNightlord:
This PR adjusts version prompts to have Bforartists instead of just Blender, still wip could use some adjustment with the wording.

Saving prompt 
image

Status bar
image

image

Resolved #5356, #5355

### DIFF
--- a/scripts/startup/bl_operators/wm.py
+++ b/scripts/startup/bl_operators/wm.py
@@ -3403,7 +3403,7 @@ class WM_MT_splash_quick_setup(Menu):
 
         layout.separator(factor=2.0)
 
-        layout.label(text=f"Bforartists {bpy.app.bfa_version_string} is based on Blender {bpy.app.version_string}") #BFA - added to first use
+        layout.label(text=f"Bforartists {bpy.app.bfa_version_string} is based on Blender {bpy.app.version_string}")  # BFA - Show blender version
 
         layout.separator()  #BFA - added to first use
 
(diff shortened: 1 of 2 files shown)


Draise14:
Looks very useful. What's pending? Only documentation?

zNightlord:
> Looks very useful. What's pending? Only documentation?

I was thinking the text could use some change so.

Draise14:
Will hold off this review after the release branch. Seems like a lot of core changes here.

Draise14:
Ok, finally, will review.

Draise14:
# Pending refinement:
1. The about page seems to be broken now. 
image

2. Remove the version + make it "Alpha" instead of "a" as it might be misunderstood. 
"Blender 5.0.0 Alpha sub 60" 
image

3. Make the splashscreen consistent with the status bar. 
image

4. Condense the overwriting:
` const char *message_line2_1 = RPT_("Saving it with this version may cause data loss.");`
Remove the second Bforartists, it is excplicit already and implied. 

` RPT_("Saving with Bforartists %s (Blender %s)"),`
Removed "based on" as it is implied by the Bforartists context already. 

5. Make it a *.blend file, it isn't technically a Bforartists file
`"Save the current *.blend file with a numerically incremented name that "`

Draise14:
@zNightlord should I give this a hand to try get the About screen to draw correctly? The rest seems like minor fixes - it's only that one really.

zNightlord:
> @zNightlord should I give this a hand to try get the About screen to draw correctly? The rest seems like minor fixes - it's only that one really.

About screen, I am not sure how that breaks, I only see that kind of behavior in Lite build since cmake config  for `WITH_BUILDINFO` is  off.

Draise14:
Ah, maybe I was in a lite, will confirm. One sec.

PR #5477, opened 2025-08-28
Fix: 5476-mac---missing-function-causing-crash-again

Blenux:
- Readding the function again, this will stop mac from crashing again.


Draise14:
Add a comment so it won't get nuked next merge!

PR #5485, opened 2025-08-29
Fix: 5484-compositor---filter---remove-sunbeams

Blenux:
-- removed sunbeam entry and icon


Blenux:
Manual already has the sunbeam info, however the information is out of order and duplicated, have a update manual for this ready to go.

PR #5486, opened 2025-08-29
Fix:5483-compositor---output---group-output-is-unknown-in-sideshelf

Blenux:
-- wrong name


Draise14:
Ah yeah, now I recall they changed the api..

PR #5490, opened 2025-08-30
Merge - Week 35

esnosy:


ReinerBforartists:
Be careful with any change at userdef_default_theme.c . This needs a task, we need to recreate the theme then. And usually it is the other way around, fix the theme, generate userdef_default_theme.c

esnosy:
> Be careful with any change at userdef_default_theme.c . This needs a task, we need to recreate the theme then. And usually it is the other way around, fix the theme, generate userdef_default_theme.c

I agree will do

PR #5492, opened 2025-08-30
Fix: Header: Remove duplicate entry of Grease Pencil Auto-Masking

### LINKED ISSUE #5491: Grease Pencil in sculpt mode, header, Double entry auto masking
One should be enough ...

Image

i-georgiev28:
Fixes #5491


i-georgiev28:
Uhh a bit funny situation but the account I sent the PR from shouldn't be used as it's only for school so I'll just redo from my other account :\

Issue #5501, opened 2025-09-06
Week 36 Extra: VSE api deprecation


zNightlord:
The addons like Power Sequencer, 3D Sequencer, Story pencil still using the old api which is removed in this [commit](https://projects.blender.org/blender/blender/commit/3978908c98e265d427142380eb18e8ba8705d1e8)

3D Sequencer part related to #5474 

<img width="1094" height="211" alt="Image" src="https://github.com/user-attachments/assets/a7b0bd13-453a-47ef-ad57-745271cab43a" />

PR #5504, opened 2025-09-07
VSE deprecated "sequence" to "strip"

zNightlord:
Resolve: #5501


zNightlord:
This can be merge before week 36

Issue #5514, opened 2025-09-08
Properties Editor - Output -  Color Management Panel - Use Curves float left


Draise14:
<img width="338" height="679" alt="Image" src="https://github.com/user-attachments/assets/84f1a778-75fd-402a-a26f-8ffb6aa29eff" />

Right now the indent is super wild. Could be improved.

Blenux:
@Quackarooni you still doing this one?, have a fix for it etc.

Quackarooni:
@Blenux Actually done with it, just forgot to make the PR. 😅

PR #5527, opened 2025-09-10
Properties Editor - Output - Color Management Panel - Use Curves float left #5514

Quackarooni:
Fixed the alignment of the `Use Curves` and `Use White Balance` properties.
Also added the `ICON_DISCLOSURE_TRI_DOWN` icon when those properties are expanded.

|  | Before | After |
| --- | --- | --- |
| Compact | image | image |
| Expanded | image | image |


Blenux:
hmm much better then my way :smiling_face_with_tear:

Quackarooni:
> hmm much better then my way 🥲

don't be too down about it, had to futz around the code for quite a while to get this. 
since the layout code wasn't immediately straightforward. 😅

Blenux:
Yeah nah all good, end of the day if it works it works haha, just cleaner.

image

Issue #5509, opened 2025-09-08
Areas - Top Left Corner Menu - Add icons and expose the toggle


Draise14:
<img width="587" height="278" alt="Image" src="https://github.com/user-attachments/assets/86922166-fb2e-4833-8914-d3584f3d7d85" />

New top left drag handle has a menu popup that needs icons. 
We could also expose the the "Hidden Editor Toggle" option to make this more discoverable (often users complain or comment that they can't find it). This means that the operator can be in a LMB menu header menu now. 

**Relevant task:**
https://github.com/Bforartists/Bforartists/pull/5500/commits/83e59eea9c72a0ec92ba2d6f89b73529c8560c9a

PR #5535, opened 2025-09-10
Areas - Top Left Corner Menu - Add icons and expose the toggle #5509

Quackarooni:
| Before | After |
| --- | --- |
| image | image |


Draise14:
Focus mode needs to be made consistent with how we have it named (old way) which is full screen I think.

Other than that seems ok.

The area I lcon could be improved (practically similar to wireframe), but that can be done another day

Quackarooni:
> Focus mode needs to be made consistent with how we have it named (old way) which is full screen I think.

This is what `View > Area` in the header shows, so I'll just match the name and icon to that.
image

Issue #5288, opened 2025-07-17
Geometry nodes - lots of missing tooltips in add menu


ReinerBforartists:
We need the name and the descriptions of the tools in the tooltip. Without it we cannot distinguish these tools in the sidebar in icon mode.

There is at least in this shown text menu lots of tooltips missing. "Add a Node to the active tree" is no valid toolname nor a description for it.

Check for missing tooltips, fix them.

![Image](https://github.com/user-attachments/assets/cf17c4aa-985c-4124-8845-8f2af406e269)

Draise14:
We reverted these tooltips to the operator tooltip instead of the RNA_nodes.cc tooltips - as they no-longer were working. So now the tooltips are drawn per node code with the title of the operator that calls that operator. We reverted to that system to use the tool name and description the Blender way, but didn't add back the double "add" title back in due to your pre-approval here: #4989

ReinerBforartists:
Yeah, but now it is missing :)

<img width="930" height="741" alt="Image" src="https://github.com/user-attachments/assets/735bcada-6111-49f3-bd8a-3c5ec1a208b2" />

ReinerBforartists:
> due to your pre-approval here: https://github.com/Bforartists/Bforartists/issues/4989

Just curious, where do you read a pre approval here? I am retired, i help out when i can. I am not into details anymore.

Draise14:
Ah, you mentioned the double title was redundant, so I removed the latter one, kept the node title based on the node name.

Draise14:
But if anything, we can comb each old tooltip and migrate them to each node title/tooltip in their respective documents if necessary

ReinerBforartists:
Yeah, we still need the title and the tooltip :)

Draise14:
Ah ok, so was a misunderstanding, and the referenced I opened previously is still relevant. I did close the other it set to a per node per task case, but seems this will need to be a parent task. 

This is out of scope for the release though, so will be for 5.0.0

ReinerBforartists:
Sorry for the misunderstanding. As told, i am not into details anymore. 

Could be something for a point upgrade ...

Quackarooni:
We could prolly have this change be done to upstream Blender, that way we don't have to add BFA comments to a bunch of individual node files. _(and worry about having to preserve those when doing merges)_

I doubt they'd say no to a PR that just adds a bunch of missing tooltip descriptions for them. 
[...]

PR #5548, opened 2025-09-12
Texture Node Editor - Add descriptions to nodes

Quackarooni:
Adds descriptions to Texture Nodes, which are missing from https://projects.blender.org/blender/blender/pulls/145772.

Related Task: https://github.com/Bforartists/Bforartists/issues/5288 
Note: This patch only partially addresses it, don't close the issue until the change from Blender main is also merged.


Draise14:
The Texture Node are not documented as they are legacy, so can close!

PR #5558, opened 2025-09-15
Fix #5554: Disable Material icon rendering causes crash through asset shelf

### LINKED ISSUE #5554: Disable Material icon rendering causes crash through asset shelf

zNightlord:
This causes asset shelf material preview display to not render and crash

Resolve: #5554


Draise14:
I did wonder about this crash, but wasn't sure what caused it.

Draise14:
Good catch.

Issue #5568, opened 2025-09-18
Brush Settings - Fix inconsistency in drawing `use_locked_size` property


Quackarooni:
Currently there are two conflicting methods of drawing this property.
One is in `VIEW3D_PT_tools_grease_pencil_v3_brush_advanced.draw` where it is being drawn as expanded tabs.
And in the other is in `brush_shared_settings` where it's drawn as a dropdown.

| Grease Pencil (Draw Mode) | Mesh (Sculpt Mode) |
| --- | --- |
| <img width="250" height="80" alt="Image" src="https://github.com/user-attachments/assets/0044b997-1e93-48a6-a9e6-53dcc832f1cd" /> | <img width="259" height="153" alt="Image" src="https://github.com/user-attachments/assets/ecc10ab8-9225-4599-a1fe-ce8bd7dd51ff" /> |

There should be consistency in how the same property is drawn, and at least to me it makes more sense to draw a two-item enum as tabs than a dropdown.

NOTE: This also affects external addons that use `brush_shared_settings`. 
One we currently know of is [Paint System](https://extensions.blender.org/add-ons/paint-system/), where there's a noticeable inconsistency between using it in Blender vs. BFA:

<img width="975" height="423" alt="Image" src="https://github.com/user-attachments/assets/17a800d1-5d82-405a-9d1e-84641451c6f6" />

Draise14:
Good catch, yeah Pink and Tawan pointed this one out.

PR #5569, opened 2025-09-18
Brush Settings - Fix inconsistency in drawing `use_locked_size` property #5568

Quackarooni:
This fixes inconsistencies between how the `use_locked_size` brush setting in drawn in different contexts.

| Grease Pencil (Draw Mode) | Mesh (Sculpt Mode) - Before | Mesh (Sculpt Mode) - After |
| --- | --- | --- |
| Image | Image | image |

This should also alleviate inconsistencies in addons like Paint System when used in Blender vs BFA.
| Blender | BFA - Before | BFA - After |
| --- | --- | --- |
| image | image | image |


Draise14:
This is where.. documentation is needed to catch these... Hmm..

PR #5579, opened 2025-09-18
Preferences - Theme - New Regions common settings: needs an icon and theme update #5561

### LINKED ISSUE #5561: Preferences - Theme - New Regions common settings: needs an icon and an update
Image

Common region areas got combined and thus nuked our theme settings (asset shelf, and not sure on the others)

This may need a pass to recompile the theme withour own. 

# Task
- Add relevant icon (MENU should do)
- Update asset shelves and other settings to old theme settings

Draise14:
Updated UV map areas to be more consistent and usable with the 3D view with white on darker background. Updated keyframes and channels to be the same
Updated panel borders, removed them
Updated common panel colors


Draise14:
Documenting.

Issue #5591, opened 2025-09-21
Mesh, edit mode, mesh menu, transform - Set dimensions addons throws error


ReinerBforartists:
Our set dimensions addon throws an error in the console and is not showing.

![Image](https://github.com/user-attachments/assets/37ec73bc-1d7c-4312-b219-5533bc11f201)

![Image](https://github.com/user-attachments/assets/92677c31-be95-4cae-9e5f-bcc6dc476c7c)

PR #5592, opened 2025-09-22
Mesh Edit Mode - Mesh > Transform - "Set Dimensions" addon operator throws error #5591

Quackarooni:
The error is caused by the draw function using a panel method in a menu, a problem caused by the `draw` function being shared by both. This patch separates them, allowing both to be drawn correctly.

image
image


Draise14:
No front end change for documentation.

Issue #5570, opened 2025-09-18
3D View - Object > Animation - Add missing "Bake Object Transform to Grease Pencil" operator


ReinerBforartists:
We have a missing entry here

![Image](https://github.com/user-attachments/assets/13f3e984-0d9e-40cf-b5dc-f84f57b80007)

ReinerBforartists:
Hm, and the object transform seems also be missing. But i have no idea how to reveal it.

ReinerBforartists:
![Image](https://github.com/user-attachments/assets/20792a80-7695-4243-bcd6-f2f3fc5eb6c6)

Draise14:
Ah nice, they added it back. It wasegacy, got nuked, then must have gotten updated.

PR #5595, opened 2025-09-22
3D View - Object > Animation - Add missing Bake Object Transform to Grease Pencil operator #5570

Quackarooni:
NOTE: Also removed the commented legacy operator, because it's no longer in Blender code.

| Before | After |
| --- | --- |
| image | image |


Draise14:
Before pulling, is this also present in the context menu and sidebar now?

ReinerBforartists:
No more quotation marks in the issue titles please. They let our automated issue generation for the manual fail.

Draise14:
If you use a regedit in the way the strings are called and replace the quotations before building the commit, it will make it more robust. This is a tall order to ask to not use a common English standard. Best to update the script.

Quackarooni:
> Before pulling, is this also present in the context menu and sidebar now?

@Draise14 Doesn't look like the case, it doesn't seem like we show the `Animation` submenu in the context menu or sidebar, thus this operator doesn't show up either.
image

Draise14:
Ah it's with the power user tools that these do get showed in the sidebar - opt in.

PR #5597, opened 2025-09-22
Default Assets - Move Shaders, Geometry Nodes, Compositor nodegroups to own library #5585

### LINKED ISSUE #5585: Default Assets - Move Shaders and Compositor nodegroups to own library
Right now the node groups of the assets could be a bit more discoverable in the add menu. We haven't had people discover these node groups often, probably due to lack of documentation, tutorials and discoverability (asset shelf helps though). So if it is it's own library they will be able to get the categories built into the other ones. 

# Rule
1. Sub categories are important to not be invasive to established library
2. Asset nodegroup type should be in it's relevant category (texture node group should be in the texture category). 

This means to make room for the Compositor assets, let's have a Compositor library.

Draise14:
Added compositor library with initial shapes from shader library
Split shader library and geometry nodes library categories to own sub-category
Updated shader library headers and properties


Draise14:
It's a pretty uninvasive commit, so will pull.

Issue #5602, opened 2025-09-23
Properties - Modifiers - Change idname of Add Modifier menu


Quackarooni:
Some addons like [Modifier List](https://extensions.blender.org/add-ons/modifier-list-fork/) expect the "Add Modifier" menu to contain all assets, as this is Blender's behavior.
We've split ours into separate menus for built-in & asset modifiers, but let the built-in modifier menu keep the original menu name: `OBJECT_MT_modifier_add`

This leads to differing behavior between BFA and Blender, as asset modifiers are inaccessible in ours if the addon is enabled.
| BFA (Built-in Modifiers only) | Blender |
| --- | --- |
| ![Image](https://github.com/user-attachments/assets/d128cd1e-2973-4b54-bc50-b5964e14f7e2) | ![Image](https://github.com/user-attachments/assets/a2077818-b6e0-4e4a-b372-f154bf8b4ea6) |

The fix I propose is to rename the built-in modifier menu as `OBJECT_MT_modifier_add_builtin` and restore `OBJECT_MT_modifier_add` to be a unified menu like Blender. We won't use that unified menu as that's not the UX we want, but we'll just have it there so that addons can point to it and not get extremely different behavior.

PR #5603, opened 2025-09-23
Properties - Modifiers - Change idname of Add Modifier menu #5602

Quackarooni:
The behavior for the Modifier List addon is now identical with base Blender, all modifiers are drawn in a single menu.

| Before | After | Blender (for comparison)|
| --- | --- | --- |
| Image | image | Image |

### DIFF
--- a/scripts/startup/bl_ui/properties_data_modifier.py
+++ b/scripts/startup/bl_ui/properties_data_modifier.py
@@ -113,8 +113,54 @@ class DATA_PT_modifiers(ModifierButtonsPanel, Panel):
         layout.template_modifiers()
 
 
-# BFA - Heavily modified to be a column menu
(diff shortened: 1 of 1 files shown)


Draise14:
EDIT: Wait, this is only to be compatible for the addon? Not core. Freaked out there a second.

Quackarooni:
> EDIT: Wait, this is only to be compatible for the addon? Not core. Freaked out there a second.

Yeah, there shouldn't be a change in behavior for when the addon isn't enabled.
It should still be like this:
image

We can't really make the addon draw separated modifier menus without contacting the developer and changing the code. So the next best option to just the make the bl_idname it's referencing point to a combined menu, since that's what it's expecting from Blender. Then we just move ours to a different bl_idname. _(This isn't a problem as nothing other than the Modifiers tab uses the menus.)_

PR #5617, opened 2025-09-26
5611 3d view   curves   sculpt mode   add tabs and tools to tool shelf

Draise14:


Draise14:
Gah, human error, it's late. Will rest.

PR #5637, opened 2025-09-30
Add node color background blend

### LINKED ISSUE #5618: Color Nodes Backgrounds - Add to preferences
Image

We could make this a user preference in the interface settings to opt-in or out like we did for the colored outliner rows. I like this a lot, I always color nodes and this helps see things at a glance, which is what we are on about. 

> In [this line](https://projects.blender.org/blender/blender/src/branch/main/source/blender/editors/space_node/node_draw.cc#L3177), replace
> 
> `UI_GetThemeColor4fv(TH_NODE, color); `with:
> `UI_GetThemeColorBlend4f(TH_NODE, color_id, 0.2f, color);`
> 
> Where 0.2f is how much you want to blend.
> 
>

zNightlord:
Further Question: Should this not be store under theme settings? but in a general user preferences?
Resolve #5618


Draise14:
Like the colored rows, I consider this an interface preference (opt in or out). The theme is about color.

Draise14:
Though adding the slider is nice. Let's keep it as an interface setting and slider so we can do unent it easier. We don't docunent theme settings.

Draise14:
image

Here would be a good spot for these settings. So it's a "feature" we can discover instead of have it deep in the theme settings.

zNightlord:
Moved out of theme
image

PR #5665, opened 2025-10-05
Fix: 4999-addon---export-uv-layout---now-a-default-addon-add-icon-to-…

Blenux:
…operator

-- added new icon SAVE_UV, based it off the other export icons.


Draise14:
Looks like the icon was not included.

PR #5725, opened 2025-10-18
Fix: 5719-splashscreen---clamp-version-number

Blenux:
-- fixed to properly show the previous version.

Screenshot_20251018_141451
Screenshot_20251018_135930
Screenshot_20251018_135718


Draise14:
Seems it should load the 5.1 preferences but refers to Bforartists 4.5.

Maybe let's just say ..

"Load previous versions preferences"

Blenux:
the first image with 4.5 was just testing to see it working.

** unless you ment this part, this would get updated to show 5.0 when we change the version number??

image

PR #5728, opened 2025-10-18
Fix: 5718-power-user-tools---remove-time-jump-operators-expose-delta-…

Blenux:
-- removed jump operators from power user tools addon, since it now added to blender/bforartists by default
-- moved delta to top level (removed separator_spacer to align props to the center, check PR images.)

**Before:**
Screenshot_20251018_195002
**After:**
Screenshot_20251018_194909


Draise14:
Nice. I would try make the delta go in-between the jump buttons and then we good. 

If possible, thinner? 

Thanks for restoring the old layout too. Almost happy to pull for that alone!

Blenux:
So put delta in the middle without text?

image

Draise14:
Yeah that would be good!

Blenux:
done.

PR #5752, opened 2025-10-22
Fix: 5744-sequencer-modifier-tab---copy-to-selected-menu-needs-icons

Blenux:
-- added icons to the Copy to Selected menu.

image


Draise14:
I think the add icon and refresh icon would do better here.

Blenux:
Done.

Draise14:
Much better!

PR #5759, opened 2025-10-23
Fix: 5738-mesh-with-hair-curves-in-sculpt-mode-hair-curve-modifiers--…

Blenux:
-- aligned the boolean checkbox left.
-- removed  SOCK_BOOLEAN && !attribute_name , since we always want the row to be true.
-- added other switch/case to align things better (change one affects another).

image


Draise14:
Really nice work here! Good catch!

PR #5770, opened 2025-10-25
Fix: 5762-bezier-curve-in-edit-mode-curve-pen-tool

Blenux:
-- align close_spline prop to left.
-- added region_type context to set row.use_property_split = False in UI and PROPERTIES only and not the tools setting header.

image


Draise14:
Got a conflict here...

Blenux:
Should be good now.

Issue #5756, opened 2025-10-22
Sculpt Mode - Box Trim Tool Group - Give property labels more space


ReinerBforartists:
The label here is cut off, and hard to read. The box behind the label takes to much space. One tool upwards the label has enough space.

![Image](https://github.com/user-attachments/assets/a447f004-63ee-4f7e-af0e-c4a850ea2ceb)

![Image](https://github.com/user-attachments/assets/74d179c2-1d44-43b9-aa7b-c0cc99772dba)

Quackarooni:
Other sculpt tools have the same issue of enum properties having the label be cut short.

<img width="255" height="180" alt="Image" src="https://github.com/user-attachments/assets/8a188666-5155-4f10-89c0-2b3e29d505b2" />

<img width="239" height="254" alt="Image" src="https://github.com/user-attachments/assets/b246e0c8-dca1-4ae6-8a65-ce0a20816b38" />

<img width="234" height="166" alt="Image" src="https://github.com/user-attachments/assets/00a9a42a-5373-49da-9b83-3b70f28c2882" />

<img width="224" height="132" alt="Image" src="https://github.com/user-attachments/assets/2f0b5521-83eb-4e71-90a0-33a25ce40aa4" />

This is due to `layout.use_property_split = False` being applied to these properties.
If that seems familiar, it's the same thing we do to boolean properties to make them float left.

For consistency's sake, I'll change this for all sculpt tools. Since rn it's quite mixed as to whether they split enums or not.

ReinerBforartists:
How shall i document an all in one task? There is a reason why i report them one by one ;)

Draise14:
I guess the pull will require making ducumentation task for each affected panel to track.

ReinerBforartists:
Yeah, and that's the problem. I know it is tempting to fix issues immediately when you see them. But we loose the whole documentation chain by that. Including the documentation in which task exactly a change happened. And what you save here with writing the fix do i pay with documentation ten times more.

Quackarooni:
Will split this into multiple issues, just gimme a sec.

PR #5774, opened 2025-10-26
Sculpt Mode - Various Tool Groups - Split non-boolean properties #5756

Quackarooni:
In the interest of not ballooning this description with screenshots, I'll just show two examples:

| Before | After |
| --- | --- |
| image | image |
| image | image |

But all other tool panels should follow the pattern of setting `use_property_split = True` for non-boolean properties.
This makes enum properties have more room for their labels. Other properties like float sliders are affected, but this actually makes them more consistent with how we handle sliders in other places.


ReinerBforartists:
Also here, please no task that fixes everything at once. This is not to document anymore.

Quackarooni:
Will split the change for each tool.

ReinerBforartists:
Thank you buddy :)

Issue #5777, opened 2025-10-26
Sculpt Mode - Cloth Filter - Make enum labels have more room


Quackarooni:
<img width="239" height="254" alt="Image" src="https://github.com/user-attachments/assets/b246e0c8-dca1-4ae6-8a65-ce0a20816b38" />

The labels get cut off, as more space is given to the dropdown than the label.

PR #5785, opened 2025-10-26
Sculpt Mode - Cloth Filter - Use split layout for non-boolean properties #5777

Quackarooni:
| Before | After |
| --- | --- |
| image | image |


Draise14:
Much tidier.

Issue #5805, opened 2025-10-29
Compositor - Sequencer Mode - Unsupported Inputs icon needs to be made consistent


Draise14:
<img width="893" height="413" alt="Image" src="https://github.com/user-attachments/assets/c2673a85-8890-477f-8e16-3027667ad70f" />

Eh.. why two different icons? Make sure the small one is used or update the other one also. 

How to reproduce
1. Go to compositor editor
2. Go to Sequencer compositor mode
3. Add a group input node
4. add a socket to the group input node

PR #5825, opened 2025-10-31
Compositor - Sequencer Mode - Unsupported Inputs icon needs to be made consistent #5805

Quackarooni:
Used `ICON_ERROR` since all the other error tooltips used that.

| Before | After |
| --- | --- |
| image | image |


Draise14:
Not sure why they used the large icon there, lol

Issue #5703, opened 2025-10-15
Grease Pencil - Edit Mode - Point Menu - Set Corner Type - add missing icons


ReinerBforartists:
Missing icons are missing.

![Image](https://github.com/user-attachments/assets/257de022-afbe-475a-a2f0-8b4204096306)

PR #5830, opened 2025-10-31
Grease Pencil - Edit Mode - Point Menu - Set Corner Type - add missing icons #5703

Quackarooni:
Registered a new `GP_CAPS_SHARP` icon since Blender apparently doesn't have an icon for sharp caps.

| Before | After |
| --- | --- |
| image | image |


Draise14:
Looks like these could go to the sidebar, they seem super useful. Will make a new task.

PR #5874, opened 2025-11-05
Cleanup: Node Editor - Toolshelf - Remove erroneous docstrings

Quackarooni:
I think this was a description from the Panel example in the code templates.
Basically just got copied over to all the toolshelf panels, even though the description isn't correct here.


Quackarooni:
This doesn't make any user-facing changes, just code cleanup.
So I didn't make an issue for this, as there shouldn't be anything to document.

PR #5891, opened 2025-11-10
Fix: 5495-3d-viewwindow---add-region-toggle-pie-to-t-and-expose-to-he…

Blenux:
-- This PR replaces the T and CTRL + T keymaps, to use the WM_MT_region_toggle_pie instead of the show_region_toolbar and show_region_toolbar operators, since they are now in the pie menu with other options like header and footer or channels, depending on the Editor. 
-- Also added icons to go with them.
-- Added view - pie menu entries for all appropriate editors.

image
Screenshot_20251110_162151


Draise14:
SO GOOD.

PR #5909, opened 2025-11-12
Fix: 5905-3d-view-sidebar-animation-tab-fix-to-camera-subtab-align-pr…

Blenux:
-- align left the Fix to Camera props.

image


ReinerBforartists:
we couuuuld indent it since it is sub props of the label ...

Draise14:
Yeah, small indent would be nice.

Blenux:
done updated image on first post.

PR #5910, opened 2025-11-12
Fix: 5741-blender-icon-on-linux-instead-of-bfas

Blenux:
-- set our Bforartists name for the app id under wayland on linux (again??, swear I did this once before???).


ReinerBforartists:
> -- set our Bforartists name for the app id under wayland on linux (again??, swear I did this once before???).

It's a trap !

Most probably slipped through with a merge ...

PR #5925, opened 2025-11-14
Merge - Week 45

esnosy:


Draise14:
Still failing, guess the master may need to be pulled into this again?

esnosy:
> Still failing, guess the master may need to be pulled into this again?

it only fails on Linux with the "clang" compiler

PR #5926, opened 2025-11-14
Change compilers from clang to gcc

esnosy:
Hopefully this fixes build failure when using clang instead of gcc

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,5 +10,5 @@ jobs:
     uses: ./.github/workflows/nightly-build.yml
     with:
       os: "ubuntu-latest"
-      c_compiler: "clang"
-      cpp_compiler: "clang++"
+      c_compiler: "gcc"
+      cpp_compiler: "g++"


Draise14:
Is there a reason to do this?

esnosy:
> Is there a reason to do this?



Hopefully this fixes build failure when using clang instead of gcc

esnosy:
> > Is there a reason to do this?
> 
> Hopefully this fixes build failure when using clang instead of gcc

to fix this
image

ReinerBforartists:
May i rise my concern? This will most probably introduce more trouble than we can eat. Blender uses clang. And so should we.

esnosy:
> May i rise my concern? This will most probably introduce more trouble than we can eat. Blender uses clang. And so should we.

I agree with you but currently the error message is not useful, it says "segmentation fault"

esnosy:
> May i rise my concern? This will most probably introduce more trouble than we can eat. Blender uses clang. And so should we.

also I just tested blender main and it has same issue, I have to report it to blender

also blender has a "WITH_CLANG" option which is not enabled in our github action

ReinerBforartists:
Reporting is always good :)
EDIT, and we should update our github action then. It is best to use the same options than Blender here.

Draise14:
Seems this may only affect Linux builds. 

Would merging this be better in the merge branch or a copy of it to see if it works?

esnosy:
> Seems this may only affect Linux builds.
> 
> Would merging this be better in the merge branch or a copy of it to see if it works?

the change is very direct and small it doesn't need a different branch
EDIT and it is done to only our github it doesn't touch any of the program code

esnosy:
Also, for what it is worth, I think  Blender uses clang only on Windows ARM and macOS, so yes, for Linux they use GCC

Blenux:
> Also, for what it is worth, I think Blender uses clang only on Windows ARM and macOS, so yes, for Linux they use GCC

They use the gcc-toolset-11 on their rocky 8 build process at least,

https://developer.blender.org/docs/handbook/release_process/build/rocky_8/

PR #5929, opened 2025-11-14
Fix: 5927-workspace-layouts---hide-the-editor-type-menus

Blenux:
-- hide some missed editors type menus that we still shown.


Draise14:
Ah, got a conflict?

Blenux:
See if that works, might need to redo again with rebase from master etc.

Draise14:
Generally a merge should hold our head on startup.blend files, so hopefully won't happen again. 

@zNightlord @esnosy in future merges makes sure we keep our changes on startup blends, if there is a task from Blender to update them, make a new task to manually update them on our side.

Issue #5943, opened 2025-11-19
Node Editor - Add & Swap Menu - Group Submenu - Fix missing icons


Quackarooni:
We used to have icons to distinguish whether groups are native, linked, etc.
See https://github.com/Bforartists/Bforartists/pull/5172.

That seems to have regressed, so they should be added back again.

<img width="468" height="341" alt="Image" src="https://github.com/user-attachments/assets/62422b02-bd2b-4b8a-b641-d9d78bd71cb1" />

PR #5957, opened 2025-11-19
Node Editor - Add & Swap Menu - Group Submenu - Fix missing icons #5943 

Quackarooni:
Also made packed assets use the `PACKAGE` icon.

| Before | After |
| --- | --- |
| image | image |


Draise14:
Does the packed icon also work in the sidebar too?

Quackarooni:
> Does the packed icon also work in the sidebar too?

Bit of a late reply, sorry. But yeah, it should work there too!

Issue #5984, opened 2025-11-26
3D View - Asset Shelf - Drag in as instance property no-onger working


Draise14:
<img width="2560" height="1009" alt="Image" src="https://github.com/user-attachments/assets/ac932ab9-dc7b-458a-a5df-8f886a6455b2" />

When you drag in a collection, with the use as instance toggle off, it still comes in as an instance. It is now broken. 

1. Open the toolshelf
2. Go to the All tab, search for the shader ball 
3. Drag it in with append on, instance off
4. Notice it is added to scene as instance

Draise14:
This got to the official release wtf. I really shouldn't have pulled the merge right before a release.

PR #5985, opened 2025-11-26
Fix: 3D View - Asset Shelf - Drag in as instance property no-longer working #5984

Draise14:
This adds in some conditionals into the drag in handler to check it's from the asset shelf property. Previously they were always set to "true" instead of set conditionally - a strange regression.

No front facing user changes to document.


Draise14:
No user facing changes.

Issue #6055, opened 2025-12-19
Compositor - Add Sidebar - Add "Creative" Tab and panel


Draise14:
<img width="1920" height="537" alt="Image" src="https://github.com/user-attachments/assets/74628eb2-0fbe-4d36-bb1a-b8a1e94c7be6" />

Ehm.. this needs to show in the sidebar. These were found in the old Color group, so they need to be pulled to a new tab.

PR #6061, opened 2025-12-27
Compositor - Add Sidebar - Add "Creative" Tab and panel #6055

Quackarooni:
Add the "Creative" category in the node toolshelf. 
| Add Menu | Toolshelf |
| --- | --- |
| image | image |

Also removed its entries that used to exist under "Filter".

| Before | After |
| --- | --- |
| image | image |

Resolves: #6055


Draise14:
Ah, conflict..

Issue #6023, opened 2025-12-15
Geometry Nodes - Add Bone Info node to Add Sidebar


Draise14:
<img width="1587" height="597" alt="Image" src="https://github.com/user-attachments/assets/f0bd6720-f589-4e9d-87b5-350aa31e3a2d" />

New Bone Info node needs documentation and adding to the sidebar.

https://github.com/Bforartists/Bforartists/pull/6018/commits/5bf8fc30ccae9dbd43128d2d5e4f56ffd0d4268c

PR #6065, opened 2025-12-27
Geometry Nodes - Add Bone Info node to Add Sidebar #6023

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

Resolves: #6023


Quackarooni:
Upon further thought, `Bone Info` not following alphabetic order is bugging me, so fixed that as well.
| Add Menu | Toolshelf|
| --- | --- |
| image | image |

Issue #6134, opened 2026-01-07
NLA Editor - Sidebar > Strip > Active Clip - Add missing label and align Sync Action Length property


ReinerBforartists:
One makes it always through ...

![Image](https://github.com/user-attachments/assets/9b1e43db-4d90-4cef-86ba-c86b772c446c)

PR #6145, opened 2026-01-11
NLA Editor - Sidebar > Strip > Active Clip - Add missing label and align Sync Action Length property #6134

Quackarooni:
Decided to just have distinct labels for the property and the operator.
"Now" doesn't seem like a clear name for what the operator does.

| Before | After |
| --- | --- |
| image | image |

Resolves: #6134

### DIFF
--- a/source/blender/editors/space_nla/nla_buttons.cc
+++ b/source/blender/editors/space_nla/nla_buttons.cc
@@ -525,11 +525,12 @@ static void nla_panel_actclip(const bContext *C, Panel *panel)
   /* BFA - Use column instead of row, so that there's enough room to add labels */
   ui::Layout &sync_col = layout.column(false);
   sync_col.use_property_split_set(false); /* BFA - Align bool property left */
-  sync_col.use_property_decorate_set(false); 
-  sync_col.prop(&strip_ptr, "use_sync_length", UI_ITEM_NONE, std::nullopt, ICON_NONE);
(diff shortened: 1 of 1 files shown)


ReinerBforartists:
It misses now the decorator. The button at the end to animate the property. If Blender has it, we should have it too ...

Blenux:
Note blender has the decorator, but it dont work at all.

Quackarooni:
> It misses now the decorator. The button at the end to animate the property. If Blender has it, we should have it too ...

To left align boolean props, we set `use_property_split` to `False`, this also has the consequence of not showing the decorator.
I can manually add it back for consistency. Do note though that as Blenux said, the decorator isn't actually functional.

Any time I attempt to keyframe this, it just errors like this:
image
Same goes for a lot of other properties in this panel actually, I can't find a situation where they don't error when keyframed.

But prolly an issue to report upstream. I don't actually know for sure if these props aren't keyframeable, but I've tried a lot and they really don't seem to be.

ReinerBforartists:
Hm. Then we should report it. 

The missing decorator just jumped immediately into my face, and so i thought i mention it :)

Draise14:
The Apply Current Length is probably designed to be.. a use one UX, not something that should change during the scene playback. So I would leave it without the decorator. Keying how it should sync would be a depsgraph and UX weirdness as.. that is like syncing a strip length on the fly in the evaluation, which.. is not desired.

PR #6150, opened 2026-01-13
Fix: Crash when rendering with Cycles using Optix with both CPU and GPU #6007

### LINKED ISSUE #6007: Crash when rendering with Cycles using Optix
**Describe the bug**
When rendering anything with Cycles (GPU Compute) with Optix as the render device, even an empty scene, or a scene with only an untextured cube, it crashes immediately with a dialogue asking to restart. This only occurs when Optix is used as render device. If I use CUDA instead there's no crash.

This doesn't happen with either Blender 4.5.5 LTS, or the latest stable release of 5.0. Both render in Cycles using Optix without a problem.

**Desktop (please complete the following information):**
Windows 11 25H2, RTX 2080 Super with driver version 591.44 

**Screenshots, Example Files**

Image
[bforartists.crash.txt](https://github.com/user-attachments/files/24063708/bforartists.crash.txt)

Draise14:
Summary: In multi-device configurations (CPU+GPU), each device gets its own BVH type:
- CPU devices use regular layouts (BVH2, EMBREE)
- GPU/OptiX devices use MULTI_* layouts that contain sub-BVHs

Root Cause: Invalid assumption that all BVHs were BVHMulti objects that could be safely cast. When using OptiX, individual device BVHs are not BVHMulti objects, causing invalid casts.

Key Fixes:
- Add proper null checks to prevent dereferencing null pointers
- Add bounds checking for sub-BVH access

I tried to keep this as uninvasive as possible and should only evaluate this more robust method when on Optix mode. This hasn't been tested on Linux yet.

### DIFF
--- a/intern/cycles/device/multi/device.cpp
+++ b/intern/cycles/device/multi/device.cpp
@@ -215,6 +215,20 @@ class MultiDevice : public Device {
 
     /* Temporarily move ownership of BVH on geometry to this vector, to swap
      * it for each sub device. Need to find a better way to handle this. */
+
+    /* BFA - handle mixed BVH types in multi-device rendering
+     * 
+     * Summary: In multi-device configurations (CPU+GPU), each device gets its own BVH type:
+     * - CPU devices use regular layouts (BVH2, EMBREE)
+     * - GPU/OptiX devices use MULTI_* layouts that contain sub-BVHs
+     * 
+     * Root Cause: Invalid assumption that all BVHs were BVHMulti objects that could be safely cast.
+     * When using OptiX, individual device BVHs are not BVHMulti objects, causing invalid casts.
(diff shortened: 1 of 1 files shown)


Draise14:
Hey y'all! Mind giving this optix multi GPU test?

esnosy:
> Hey y'all! Mind giving this optix multi GPU test?

I only have Intel integrated (doesn't even support oneAPI) graphics in laptop, and AMD on PC away but I read the code and it seems ok, it builds too, so that's good, I would ask the original poster of the issue to test it by building the branch on Windows.

esnosy (approved):
Looks good to me

Draise14:
Eff it. If it causes issues we can revert.

PR #6162, opened 2026-01-15
6156 properties editor   object properties with eevee   visibility panel   indent props

Blenux:
- indented.

image


Draise14:
Same here, a little BFA comment helps rebase easier in the future.

PR #6163, opened 2026-01-15
Fix: 6147-freestyle-modifiers---align-props-left

Blenux:
- align left

image


Draise14:
In a rebase this may get missed, so don't forget the tag BFA on next!

PR #6167, opened 2026-01-16
6158 3d view item tab transform panel   refactor mode property

Blenux:
- worked out how to make the decorator work with the prop but separated, so the blank icon and 4L toggle can be before it, so it matches the surrounding layout patterns.

- also makes the properties space python version the same.

[Screencast_20260116_222425.webm](https://github.com/user-attachments/assets/938895c5-a7af-4e27-a051-096aadd0208e)


Draise14:
Much easier to keyframe when dragging down now, nice. Well done!!!!!!!!!

Issue #6193, opened 2026-01-21
Files imported from Blender don't work correctly


esnosy:
here's the fix

<img width="1423" height="195" alt="Image" src="https://github.com/user-attachments/assets/dd605488-1ff4-451e-be0d-752148575c04" />

ReinerBforartists:
This will make us lots of fun, we had it before, your fix will break our file compatibility ...

esnosy:
> This will make us lots of fun, we had it before, your fix will break our file compatibility ...

but it is correct to be able to open Blender files, our files can be resaved with correct editors, users can also resave

ReinerBforartists:
Yeah, but we cannot load our old files anymore correctly.

Draise14:
Not sure this will break anything? Worth a test.

ReinerBforartists:
Dig for it, as told, we had it before.

esnosy:
> Dig for it, as told, we had it before.

you are correct it will make the toolbar be a spreadsheet editor instead, but that's not a fatal failure, better in my opinion to be able to load both Blender and newly saved BFA files

Draise14:
Though Blender blend files can be loaded perfectly fine we have Load UI off already. 

And Bforartists blends default to info editor if load UI is on. 

I wasn't aware if a crash with file compatibility.

ReinerBforartists:
I still wonder if there is a way to introduce some fake editors, add a few more numbers to the enum, and set our editor to maybe the value 30. that way blender can introduce more new editors until we get hit by the problem again.

esnosy:
Good idea will do later

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

Draise14:
Projects Editor is slated, so yeah, good call to bump up a chunk.

ReinerBforartists:
Just test it enough. Not that this stikes back ...

PR #6196, opened 2026-01-21
Fix: Files imported from Blender don't work correctly #6193

esnosy:


Draise14:
Will this break existing startup blend files?

esnosy:
Yes

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

Draise14:
Ah, so this patch will need work with the updated Startup.blend files. So can't accept it right now. 

It also should probably get another higher enum to give head to the new Projects editor coming this year.

esnosy:
> Ah, so this patch will need work with the updated Startup.blend files. So can't accept it right now.
> 
> It also should probably get another higher enum to give head to the new Projects editor coming this year.

Done, I checked all startup and the toolbar is there, I bumped up the value to 40

PR #6208, opened 2026-01-23
Merge week 4

esnosy:


esnosy:
pulled more commits from Blender hopefully this fixes it, if not, then it would need more work

PR #6246, opened 2026-02-01
Fix: 4142-toolbar-editor---make-preferences-the-same-as-the-fixed-top…

Blenux:
- Updated the Space Toolbar Editor's Toolbar Manager to be the same has Topbar version.

- This is a temporary solution to just get the two matching for now, im in the process of refactoring/cleanup with a better unified system, which will reduce code duplication etc.

[Screencast_20260201_131243.webm](https://github.com/user-attachments/assets/fd0f9334-2cb4-40d1-94d1-7e45e2ed6fa3)


Draise14:
Looking good. Consistency is great here.

PR #6283, opened 2026-02-06
UI - Core - Make boolean properties always left-aligned

Quackarooni:
Always make booleans align left regardless if layout was split or not.
A great deal of regressions we get on our end is having to deal with booleans using a split layout in Blender main code, while non-split in ours. We save a lot of maintenance overhead on our part if we just enforce this as global behavior in `Layout::prop`.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/source/blender/editors/interface/interface_layout.cc
+++ b/source/blender/editors/interface/interface_layout.cc
@@ -2111,8 +2111,17 @@ void Layout::prop(PointerRNA *ptr,
       }
     }
     else {
-      Layout *layout_split =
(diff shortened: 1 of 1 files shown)


Quackarooni:
This should get rid of any need for the following:
- `use_property_split = False` to left-align booleans
- `use_property_split = True` to put labels outside of number sliders
- explicit calls to property decorators because non-split layouts don't support them

Draise14:
Will have to compile this one to see if the indents we do are kept.

Draise14:
eff it.. les go. Might save us work. Closes 3 tasks already.

PR #6347, opened 2026-03-10
Fix: 6345-file-browser-keymap---replace-tool-props-keymap-with-tilde-key

Blenux:
-- This simply replaces the file browser tool props toggle, for the view pie menu (since #6344)

(Accidently pushed this to #6344, but reverted it)

### DIFF
--- a/scripts/presets/keyconfig/Bforartists.py
+++ b/scripts/presets/keyconfig/Bforartists.py
@@ -1974,10 +1974,10 @@ keyconfig_data = \
        ],
       },
      ),
-    ("screen.region_toggle",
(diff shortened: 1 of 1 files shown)


Draise14:
Tilde doesn't exist worldwide, so can't test. Which pie menu is this one?

Draise14:
In this case, removing the hotkey for the toggle should suffice.

Blenux:
Are forgot about that, the same key that is in 3d view for the front, left, right, top, bottom pie view menu etc

Blenux:
Actually, just realized that the keymap was CTRL + T, however will still remove it since the T keymap now does this (duplicate keymap in a sense).

need to do the macos keymap also!

PR #6348, opened 2026-03-11
Fix: 6328-import---drag-and-drop-menu-for-svg-needs-icons

Blenux:
-- added menu icons for the drag drop menu selection option.
-- added get_icon_value function (read below)

(This is done in via addon_utils, so it can be imported for other file handlers in future when/if needed, this function is needed, since file handler rna requires a integer, and also when we add more or remove icons, their ids can change, this just automatically finds the icons integer id number.)

image


Draise14:
Clever. I will have to dig the other import multi choice icons.

PR #6350, opened 2026-03-11
Fix: 6319-3d-view---grease-pencil---draw-mode---erase-brush---add-ico…

Blenux:
-- added icons

image


Draise14:
Probably should make all the GP icons consistent... will make a task for that. But good nuff for now.

Issue #6336, opened 2026-03-05
Grease Pencil - Draw Mode - Add "sync" surface distance with brush radius


Draise14:
<img width="1605" height="787" alt="Image" src="https://github.com/user-attachments/assets/d0d2df36-8814-46e2-abed-bda2cd832ced" />

Right now the biggest issue with grease pencil drawing on surface is that it doesn't respect the distance by radius. It currently is decoupled, meaning grease pencil often "clips" on the surface of objects. Not only does this look bad, but it's very inconvenient and makes many rendering issues. And when you convert GP to mesh or tubes, it will "clip". 

It would be awesome if a user can "opt in" to sync radius and surface distance with a single button or operator, or always sync these properties when a "mode" is on. 

Bonus: Even better, if it were on a sync mode, it would respect pressure on surface too.

Blenux:
This?, no idea how to use the tool properly though 😄 

https://github.com/user-attachments/assets/97af62ea-b524-436f-8456-bb7ee4e74dbb

Draise14:
Wow that prototype was quick! How?!

Yeah that idea of it syncing is it! I can test on my tablet.
The next trick is it will sync with pen pressure too? But this sheet is a step forward. 

# To test:

1. Make a plane and a cube scene
2. set mode to draw on surface and sync
3. Draw in cube edges and scribble on the surface at different widths
4. Use the curve to tubes asset from the shelf in object mode
5. Check clipping, or render frame

Blenux:
I will push it to the branch so you can test, no idea how to use this feature fully lol and I don't have a tablet to test on anyways.

Blenux:
Do the plane and cube need to be a grease pencil??? @Draise14

Draise14:
Would be a mesh, the ndraw the Greasepencil on surface mode. 

I have started trying to compile this build, will check tomorrow.

Blenux:
ah need to do another push to this, missed some needed changes on the push!

Blenux:
should be good now (if the feature actually works lol)

Draise14:
Compiling.

Draise14:
https://github.com/user-attachments/assets/8754db85-3653-4b88-854a-9670e398abee

Does the trick. Pressure sensitivity does _**not**_ sync though. 

But I would say for now, for solid stroke with no pressure, it's a huge win!

Draise14:
[...]

PR #6358, opened 2026-03-16
Grease Pencil - Draw Mode - Sync Pressure and Radius (with additional offset) #6336 

Draise14:
image

How to use:
1. Turn on Surface drawing mode
2. Turn on Surface Sync
3. Draw

How It Works:
- When you set brush size to 1m, the surface offset automatically becomes 1m
- The extra offset (0.01m by default) is added on top
- Total offset = brush_size + extra_offset
- Pressure changes affect both brush size and total offset proportionally


Draise14:
There are some perspective issues with the drawing in normals and low angles on normals, but this is true with or without the new system.

Issue #6380, opened 2026-03-20
Toolbar/Topbar - Manager Panel - Make different functions more distinguishable


Quackarooni:
The UI/UX for the Toolbar/Topbar is not great IMO. 
What struck out to me when I first tried using it was how unintuitive the layout is.

<img width="646" height="662" alt="Image" src="https://github.com/user-attachments/assets/e9bdbc7d-65ef-457a-b498-d3e075132c8f" />

A big problem is how a lot of different things use the same button presentation. Toggles, enums, and operators are all laid out the same way, without much guidance to the user what each section does. 

While "Defaults" do an okay job indicating that section is for resetting to default settings, "Types" and "Options" do aren't as helpful, needing you to hover for a tooltip to explain their functions.

### Proposed Fix
How I envision this is putting the toggles and enum in the same column, which the enum displaying the names and the toggles presented as icons. For the reset operators, I would like if those had icons to indicate they're for resetting to defaults.

Blenux:
Sad Face.

Yeah this was a short coming when I created this, I do want to redo this, plus reducing the amount of duplicated code between Topbar and Space Editor's Toolbar similar to your #6372 issue

Any recommendations would be appreciated.

Blenux:
Something like this be better? (Toolbar Controls)

https://github.com/user-attachments/assets/7bede575-188e-40ac-ab2d-b9b1dab2c426

Blenux:
<img width="436" height="172" alt="Image" src="https://github.com/user-attachments/assets/2832570d-a85f-4ef8-8852-484bfe9f265d" />

Quackarooni:
@Blenux I was in the middle of writing a patch for this. 😅
Hence me making an issue, didn't wanna complain about something I couldn't fix.

This is what I have so far.
<img width="489" height="256" alt="Image" src="https://github.com/user-attachments/assets/4500a796-243c-4005-8264-c45a563e9ac9" />

So pretty similar to yours, though I think making it one column is nicer as you can click drag to activate/deactivate multiple sections quickly. (For the topbar, at least. Toolbar uses operators.)

Blenux:
Haha all good, yeah that does make sense actually, just used to the grid setup 😄 , the eye icon does look better!

and yeah the space toolbar is different since it's a editor type, not a global bar, been off and on ways to improve it, and so that each space toolbar editor can have its own set toggles like how other editors do, currently toggling one does it for all other space toolbar editors.

[...]

PR #6382, opened 2026-03-20
Toolbar/Topbar - Manager Panel - Make different functions more distinguishable #6380

Quackarooni:
Rework the Toolbar/Topbar to better present its properties.
- Turned the enum into a vertical list.
- Turned the toggles as side buttons with icons HIDE_OFF/HIDE_ON.

Also, on the technical side, the draw function has been split up into multiple subfunctions. This is to make it easier to implement other upcoming refactors.

| Before | After |
| --- | --- |
| image | image |

For reference on how quick it is to edit sections, here's a video of the topbar in action.
Note: The toolbar can't do this drag-behavior as the toggles are operator buttons, not props.

<video src="https://github.com/user-attachments/assets/e15113b2-6fff-454a-a735-af55d2859a4c"></video>

Resolves: #6380

### DIFF
--- a/scripts/addons_core/bforartists_toolbar_settings.py
+++ b/scripts/addons_core/bforartists_toolbar_settings.py
@@ -54,25 +54,25 @@ class BFA_OT_toolbar_settings_prefs(AddonPreferences):
     ])
 
     bfa_topbar_types: EnumProperty(name='Topbar Types', description='', items=[
-        ('Files', 'Files', 'Files Options', 0, 0),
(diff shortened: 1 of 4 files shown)


Quackarooni:
@Blenux Added you for review, feel free to give me your thoughts if the way I did things is appropriate.

@Draise14 Ultimately your call, but I worry a bit about merging a change like this so close before a release, so would advise that this could wait, along with other toolbar/topbar changes I plan on making.

Draise14:
I like the new proposal, more explicit and easier to use. Will take a bit to document! Will pull it into the release.

Issue #6321, opened 2026-03-05
3D View - Sculpt/Object Mode - Make Add Modal Primitive top level panel with updated icons


Draise14:
There is a modal primitive submenu now for Sculpt mode and Object mode. They need icons and probably some consistency with discoverability and accessibility, by being a more top level kind of set of operations. 

If a developer misses these use cases, they are not obvious enough. *cough* me *cough*

# Tasks

<img width="789" height="608" alt="Image" src="https://github.com/user-attachments/assets/07afbfbc-12c8-4a1c-a2e5-5022e946c1ab" />

1. Remove from header in Sculpt, these are redundant
1. Add icons in our style, with the plus sign to show the modal nature of. These are generated from the toolshelf.blend file
2. Make the add sub-menu go consistently to the top panel level in it's own panel either in it's own tab or additionally in the Tool tab. 

https://github.com/Bforartists/Bforartists/pull/6359/changes/81e41f5f73fa09ed4a7e18248c292a32bebed97f

PR #6386, opened 2026-03-21
3D View - Sculpt/Object Mode - Make Add Modal Primitive top level panel with updated icons #6321

Quackarooni:
image

- Removed "Add Primitives" menu from header
- Moved all "Add Primitive" tools to top level in toolshelf. (in all modes).

### All configurations
| | |
| --- | --- |
| image | image  |
| image | image |

### DIFF
--- a/scripts/startup/bl_ui/space_toolsystem_toolbar.py
+++ b/scripts/startup/bl_ui/space_toolsystem_toolbar.py
@@ -3915,7 +3915,7 @@ class VIEW3D_PT_tools_active(ToolSelectPanelHelper, Panel):
         'OBJECT': [
             *_tools_default,
             None,
-            *_tools_view3d_add,
+            *_tools_view3d_add, # BFA - Expose at top-level
         ],
         'POSE': [
             *_tools_default,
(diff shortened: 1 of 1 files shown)


Quackarooni:
Took me a long time to figure out how to get the icons to export from the .blend correctly. 😅

Just did the simple thing here of reusing our icons for primitives, but scaled up.
I kept the plus part of the icon as it is, feel free to change that as fit.

Also, we have multiple copies of `toolbar.blend` in the repo, I just changed both the ones under `bfa-asset` and `release`. _(`release` is the one used by the exporting script)._

Draise14:
I should upload my documentation for that. 😅 got some local.

Nicely done though. 

About exposing them top level, in another panel could be preferable if people do want to collapse it. But will need to be tested once tabs goes off. Though now I'm not sure how to do that with modal operator buttons. 

Did you manage to do the same for the Object mode?

Draise14:
If it can't be placed in a panel, drop to last of the list. Brushes first.

Quackarooni:
@Draise14 
> About exposing them top level, in another panel could be preferable if people do want to collapse it. But will need to be tested once tabs goes off. Though now I'm not sure how to do that with modal operator buttons.

Don't yet know how to separate them in a different panel. The code for the tools panel is obscure and it kinda uses some funky stuff with decorators and all.😅

>Did you manage to do the same for the Object mode?

Yeah, it's done in all modes. (Object, Edit, Sculpt)

PR #6390, opened 2026-03-23
Node Editor - Tweak various Add Menu entries

Quackarooni:
- Enforce alphabetical order
- Remove Get/SetGeometryBundle nodes (still experimental, not good to expose)

### DIFF
--- a/scripts/startup/bl_ui/node_add_menu_compositor.py
+++ b/scripts/startup/bl_ui/node_add_menu_compositor.py
@@ -284,8 +284,8 @@ class NODE_MT_compositor_node_utilities_base(node_add_menu.NodeMenu):
         del context
         layout = self.layout
         self.draw_menu(layout, path="Utilities/Math")
-        self.draw_menu(layout, path="Utilities/Vector")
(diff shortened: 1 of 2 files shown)


Draise14:
If things are not hidden by an experimental toggle by defualt, Bforartists tries to never have hidden elements in the GUI. If not, it can lead to it going missing in future merges.

If it's unstable, that's on Blender. It should also be safe to exoose since it's mostly a power user function.

Quackarooni:
@Draise14 Oh, those are also already exposed under Geometry Read/Write behind an experimental toggle.

image
image

I was guessing that duplicating them under Utilities > Bundle was a convenience feature. *(Those entries were introduced in your new Bundle nodes commit.)*

I don't think it's time to do that yet, since if Blender moves it out of experimental or do something like change node names, it might be easy to miss we have duplicated entries in this menu. Better to do that once the nodes have been finalized so we don't have to worry about keeping them in sync.

Draise14:
Oh wow. So it is experimental. My bad. Will pull in once I'm at the computer.

PR #6400, opened 2026-03-24
Topbar - Fix typo on startup file message

Quackarooni:
Say "Topbar" instead of "Toolbar".


Quackarooni:
This should've been in the other PR, but I forgot to push to origin, and only realized when I woke up. :'D

PR #6413, opened 2026-03-28
Toolbar - Manager - Directly show toggle properties instead of using operators

### LINKED ISSUE #6383: Toolbar - Manager Panel - Pressing ENTER disables toolbar sections
Pressing ENTER while the manager panel is open will disable the first active panel section. 
Repeatedly doing this will go through all the sections and disables them one by one.

https://github.com/user-attachments/assets/485e48c2-7e8e-4441-ab66-5ad4f05e4bfc

This behavior is not present in the topbar where the toggles are props, not operators.
The cause is likely using `.active_default` to display the operator button as pressed.

Image

Something potentially worth exploring would be using `.operator(..., depress=True)` instead.

Image

Quackarooni:
Directly draw the show/hide toggle properties in the Toolbar manager, in a similar manner to the Topbar. This allows the user to quickly set multiple toggles via click-dragging rather than pressing one-at-a-time.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4904cdd8-6719-47c3-937c-9ad6f5de183b"></video> | <video src="https://github.com/user-attachments/assets/a060409f-1da6-4daa-a218-631bfe33455d"></video> |

The reason operators were used instead of properties was because the toolbar did not update when a property was toggled, because it doesn't trigger an area redraw like an operator does. The same result can be achieved by adding an `update` callback to the properties.

Also resolves #6383.

### DIFF
--- a/scripts/startup/bl_ui/space_toolbar.py
+++ b/scripts/startup/bl_ui/space_toolbar.py
@@ -10,6 +10,10 @@ from bpy.types import (
 )
 from bpy.app.translations import contexts as i18n_contexts
 
+from bl_ui.utils import (
+    icon_button,
+)
+
 ######################################## Toolbar ########################################
 
 class TOOLBAR_HT_header(Header):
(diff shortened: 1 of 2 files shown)


Quackarooni:
@Blenux Tagging you just to confirm if the reason for using operators was to trigger a redraw?
And if the alternative approach taken by this PR is sufficient, or if I'm missing something else?

Blenux:
Will have a looks.

PR #6442, opened 2026-04-03
Fix: origin/6441-space-toolbar---wrong-align-icon-names

Blenux:
-- ALIGNX = ALIGN_X (was ALIGNHORIZONTAL previously) 
-- ALIGNY =  ALIGN_Y (was ALIGNVERTICAL previously)

This matches the other places its used at.


Draise14:
Ah woops, pushed this with a typo

PR #6446, opened 2026-04-04
Fix: 6445-topbartoolbar---add-region-hud-support

Blenux:
-- added region hud support for topbar/space toolbar areas

> This simply adds the Region Hud/Adjust Last Operation to the Topbar and Space Editor Toolbars.
> 
> This simply allows the user to adjust the settings for that object, just like in the Editors - Add Menu.
> 
> https://github.com/user-attachments/assets/8dc39272-5297-4c74-88e9-81c126e9fd1b


Draise14:
Nicely commented.

Issue #6437, opened 2026-04-03
Properties - Sequencer Strip - Transform - Mirror needs indent to new line


Draise14:
<img width="1934" height="854" alt="Image" src="https://github.com/user-attachments/assets/e76383d7-e5ef-4126-a806-e291c28ac591" />

PR #6453, opened 2026-04-07
Properties - Sequencer Strip - Transform - Mirror needs indent to newline #6437

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

Resolves: #6437


Quackarooni:
I only fixed the alignment issue, but an alternative here is to let the mirror axes be checkboxes instead of toggles:
image

@Draise14 Any thoughts on whether these should be checkboxes or buttons?

Draise14:
I think consistency with the other mirror panels should need be considered. As far as I remember, they are buttons.

PR #6552, opened 2026-05-16
Refactor: Toolsystem - Sidebar Panels - Complete rewrite of drawing code

Quackarooni:
The operators on the left sidebar can either be presented as a column of buttons with labels, or a grid of icon only buttons.
Not only were these separate implementations, but also, the one column, two column, & three column views of the icon buttons were separate. This means that each panel has to define 4 distinct ways of drawing the buttons on the screen.

```python
column_count = toolsystem_column_count(context.region)

if column_count == 4:
    # Draw as text buttons
elif column_count == 3:
    # Draw as icon buttons (3 columns)
elif column_count == 2:
    # Draw as icon buttons (2 columns)
else:
    # Draw as icon buttons (1 column)
```

Not only does this result in a lot of code, but it also adds a significant burden when making any additions/changes to the operators in these panels. It also leaves a lot of room for potential errors where not all drawing implementations got updated correctly.

The changes here aim to solve those issues by making it so that panels only need to define an `entries` list to enumerate the operators/menus they need to draw, and the drawing logic is handled by separate helper functions. 

```python
entries = (
    # Operators/menus to draw.
)

draw_entries(layout, entries, context)
```

This makes it easier to add, remove, and re-order different buttons on these panels.
It also adds the assurance what is drawn is consistent across all the different views.

In terms of user-facing changes, in theory this shouldn't make any difference to what the user sees.
In practice, there might be slight changes where the old implementation was inconsistent in how it drew the different views. Most of notable of this is that some panels in the three and two column view skipped drawing separators, where this new implementation doesn't.


Quackarooni:
Another fat PR, basically the spiritual successor of this one: https://github.com/Bforartists/Bforartists/pull/5315 😅

Draise14:
Holy, it nuked 4000 lines. I will have to compile this to see.. but sounds very promising!

Make sure the maintenance and how to update this is heavily commented.

Quackarooni:
> Make sure the maintenance and how to update this is heavily commented.
I added comments where I found sensible. I think which helper classes to use is clear from the names, and how to use them is similar enough to using the `layout.(...)` functions that it doesn't need specific explaining.

This code file doesn't have anything that needs manually adjusting, unlike with the node shelf that had manual padding.
There is one caveat where `layout.operator_enum` isn't possible to support with the new code.
Added a comment explaining this in the function that's preventing it from being doable.

Draise14:
So for enums to be exposed in the shelf we would need to make a ghost python handler for the enun like in the old days?

Draise14:
Things to test: 
1. If they add to the quick menu right right click
2. If hotkeys can be assigned to them from the context menu

Quackarooni:
> So for enums to be exposed in the shelf we would need to make a ghost python handler for the enun like in the old days?

Not sure what you mean by that, but I don't think so. We don't really expose any regular enums in this part of the UI, so I don't think that would be a concern.

What I was talking about were operator enums, they're a layout function that basically automatically handles drawing all possible options of an operator based on an enum prop. 

```python
entries = (
    # Instead of layout.operator_enum("object.convert", property="target")

    # We have to individually list each possible configuration for them to be properly drawn in the icon grid.
    OperatorEntry("object.convert", text="Mesh", icon='OUTLINER_OB_MESH', props={"target": 'MESH'}),
    OperatorEntry("object.convert", text="Curve", icon='OUTLINER_OB_CURVE', props={"target": 'CURVE'}),
    OperatorEntry("object.convert", text="Curves", icon='OUTLINER_OB_CURVES', props={"target": 'CURVES'}),
    OperatorEntry("object.convert", text="Point Cloud", icon='OUTLINER_OB_POINTCLOUD', props={"target": 'POINTCLOUD'}),
    OperatorEntry("object.convert", text="Grease Pencil", icon='OUTLINER_OB_GREASEPENCIL', props={"target": 'GREASEPENCIL'}),
)
```
The main problem here is that it tries to squish all buttons in a single cell of the icon grid, not put one button for each cell.

Note: We already do this in the old code, if the column count is 2 or 3. It's an inherent limitation of how we draw buttons in a grid. We also do this in cases where we want to customize the labels/icons, since `layout.operator_enum` doesn't let you control that.

So it's not an unusual thing to be doing in this part of the code actually.

Quackarooni:
> Things to test:
> 
>     1. If they add to the quick menu right right click
> 
>     2. If hotkeys can be assigned to them from the context menu

Tried both on my machine, seems to work fine. But do double-check it on your end if you wanna be sure.

Blenux:
Very nice, I had something very similar to this locally, but has always got distracted with other task/things.

Draise14:
Will compile this next week, and may require merge fix. So after the 5.1.2 release for the 5.2.x era.

Quackarooni:
Linux build failure seems to be from lfs

Draise14:
Ah probably.

Blenux:
I check this out, all seems to be working fine, awesome work!,

we do loose the right click - edit source, that takes you to that operator line directly in the old way, but way less code is better to have! 

I also reran the Linux check yesterday, all passes now.

Draise14:
Looks like we have done some work to the sidebars that may need to either be merged in or updated. Will pull this in after! For the 5.2.0 release.

Draise14:
Will pull after the merge.

Draise14:
Ok! Guess we can merge this in now.

Quackarooni:
@Draise14 If it's fine, I wanna do a quick check beforehand later today. Just wanna make sure that none of the tools sidebar related bug fixes we did got accidentally reverted when moving to the new system.

Draise14:
Sure! Polish away. A lot has happened these weeks yeah.

Quackarooni:
@Draise14 Should be done now, as I've suspected, most of the previous bug fixes got lost when merging with this new system. So I've added them back.

Also missed the fact that the panels for the UV Editor are in another file, updated that one too now.

PR #6585, opened 2026-05-29
Sidebars - Compact Tabs option makes columns too thin, offsetting icons

### LINKED ISSUE #6518: Sidebars - Compact Tabs option makes columns too thin, offsetting icons
Image

The sidebar categories change their width when in the new "Compact" mode, which means our columns need to figure out a better way to consider this conditional.

Blenux:
- Introduced a new function to preserve column count when toggling toolbar tab visibility.
- Updated the snapping logic in the tools region to account for tab visibility and compact mode.
- Enhanced the initialization of toolbar widths during workspace changes and editor initialization.
- Removed obsolete toolshelf tab width management code from the image editor and other spaces.
- Added support for hiding category tabs in specific regions, improving UI flexibility.
- Updated RNA properties to manage toolbar tab visibility across different spaces.
- Ensured toolbar widths are recalculated after loading files and during startup.

@Draise14 - This was bit more then the original branch intended, but needed to be done and been wanting remove my older setup I have and just refactor it to get it out the way, tried to break it but seems solid enough (will see though!).

@Quackarooni - This has some minor changes to your toolsystem_column_count helper function, mind checking if all is good, all for making this better has possible.


Draise14:
Had a quick compile and look. Looks green enough to me, better than previous versions for sure. So I will pull in. If we find anything, we can then open up a new task.

PR #6595, opened 2026-05-30
Properties Editor - Strip Modifiers - Add a "quick open compositor....

### LINKED ISSUE #6579: Properties Editor - Strip Modifiers - Add a "quick open compositor window" to button/operator AND/OR make a Sequencer compoitor workspace already setup
Image

# Problem
Right now when you use compositor assets in the sequencer, there is no easy connection or way to have a user connect the compositor editor to the sequencer strip, it's possible, just 3 steps of know-how to get there. You can pick an asset, but not get to it quickly - neither from the Strip Modifier stack, nor as a Workspace. 

# Task
1. Add a "quick open compositor window" to a button/operator either top level or in the drop down. 
2. Make a Sequencer compositor workspace already setup in the default templates for discoverability sake

Blenux:
-- Added Open Compositor Editor Window Operator
-- Added Compositor Workspace to Video Sequencer Template (wasn't in video at the time)

[Screencast_20260530_145842.webm](https://github.com/user-attachments/assets/648f4c1f-5bb2-4ca6-a50a-a79dd25267d4)

Task #6579

### DIFF
--- a/scripts/startup/bl_app_templates_system/Video_Editing/startup.blend
+++ b/scripts/startup/bl_app_templates_system/Video_Editing/startup.blend
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e6d7e74262d9aa08bfe807321d38c40bbdb12053ed7f565b243da3d83f30bedc
-size 517578
+oid sha256:8a1c2f40fae83617d9db99af6b6d8a74201d6f48b3dcd54e665fea8031ce53eb
+size 552425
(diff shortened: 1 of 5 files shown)


Draise14:
Nice! Come to think of it, if we add this, we will also need to do it for the shader materials, geometry nodes modifier stack, and world shader and texture nodes area... 

And then the design of the button... 

Also do you think this should be too level? Maybe a small square in header? I'm the drop-down options?

Blenux:
Yeah can add it to them for consistency.

Regarding Top level, not sure with that, the modifier headers are cramp enough has is, hench why I just put this in the compositor modifier since it's only needed then.

Draise14:
What about end of modifier or in the modifier sub menu top right? 

So UX is like this
1. Tune modifier stack
2. Need more control
3. Hit button bottom of stack
4. Editor opens.

Blenux:
Will hope on soon and see what works best.

Blenux:
This is ready for merging now.

PR #6598, opened 2026-05-31
Properties Editor - Strip Modifiers tab

### LINKED ISSUE #6573: Properties Editor - Strip Modifiers tab - Split Modifier Assets like in Geometry Nodes Assets modifier stack
Image

To be consistent, we probably should have the compositor modifier for the Sequencer to use the same system as Geometry nodes, so we have two search menus and one level less deep to add compositor assets directly onto strips. 

Like in Geometry Nodes:
Image

**Relevant Task**
https://github.com/Bforartists/Bforartists/pull/6568/changes/a1f0e91f95858347e61ed3f672c628e168f8e596

Blenux:
-- Separated the Assets to their own menu just like the Geometry Nodes

image
image
image

### DIFF
--- a/scripts/startup/bl_ui/properties_strip_modifier.py
+++ b/scripts/startup/bl_ui/properties_strip_modifier.py
@@ -5,8 +5,12 @@
 import bpy # BFA
 
 from bpy.types import (
+    Menu,
+    Operator,
     Panel,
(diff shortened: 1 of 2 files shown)


Draise14:
Thank you.

Issue #6525, opened 2026-05-08
File Browser - Expose hotkey only selection and walk operators to the header


Draise14:
<img width="1490" height="624" alt="Image" src="https://github.com/user-attachments/assets/5bf3f2f8-2725-45bf-8a0b-231fdf62edf0" />

We have a bunch of hotkey only operators, they are not discoverable in the interface in the file browser. 

We should make them discoverable like we did for the Cursor in the text related modes and editors. 

# Tasks
- Add UP, DOWN, LEFT, RIGHT move selection
- Add Select UP, DOWN, LEFT, RIGHT while holding modifiers
- Add Select mouse operations for discoverability

Basically make these discoverable:
<img width="1170" height="711" alt="Image" src="https://github.com/user-attachments/assets/90240334-87ea-4d5d-9422-70079b31b60b" />

Here is what we do with others:
<img width="598" height="258" alt="Image" src="https://github.com/user-attachments/assets/b128967e-4cd0-46c2-89b4-e94c0219cc94" />
<img width="602" height="267" alt="Image" src="https://github.com/user-attachments/assets/4a9c24b0-e9c2-49cc-8836-e29d91a2fa8d" />

PR #6612, opened 2026-06-02
File Browser - Make a new operator to move to start or end of the fol…

### LINKED ISSUE #6526: File Browser - Make a new operator to move to start or end of the folder with HOME / CTRL+HOME
Image

In the file browser, you can't go to start or end of a folder easily. In everywhere else with cursor based workflows, you can (text editor, info editor, etc).

Would be nice to have these as operators. Could be done in python I think, in this case could be built into the Power User Tools addon.

Blenux:
-- Added Navigation to the File Browser header
-- Added To Top/To Bottom Operators
-- Added keymaps for To Top/To Bottom operators to all keymap presets.

Image

https://github.com/user-attachments/assets/bd49fcd8-a082-49a2-b752-c88173a40d50


Draise14:
We also have "walk" hotkeys and selection hotkeys. Should we also expose those? Left right, up down, etc.

Blenux:
Yeah seen the task for that, can do that seperate today

Draise14:
Been mulling this over. We have the first entries (back, forward, up, refresh) in the header tools already, so these would be redundant operators in the navigation menu. 

So for now I think, we try not do this if the operator is already in the header. Only exception is for the toolshelf. So we should only add the walking hotkeys (left, right, etc) like the cursor hotkeys of the text editor. Not repeat ones already in the header.

Blenux:
Header has in the Filepath bar ??

image

Blenux:
Current setup of this PR (Navigation - To Top/To Bottom) and task #6525 (Select Up/Down/Left/Right)

Navigation: Directory Navigation
Select: File/Folder Selection

[Screencast_20260603_122143.webm](https://github.com/user-attachments/assets/bba69a17-170a-48dc-bb75-c24dac3cc328)

Note: Can't use the text editor  ones since they dont work in the file browser context.

Draise14:
Ah discoverable now. And the new to top or end is nice.

Not sure what I was talking about, will pull in and document soon.

PR #6647, opened 2026-06-12
Sequencer - Color Strip - Add Width and Height to sidebar

### LINKED ISSUE #6635: Sequencer - Color Strip - Add Width and Height to sidebar
Image

I think we may need to open a debate if we should have two strip properties, since it is very disconnect with another editor which breaks on full screen. 

But alas things need to mirror and then be documented for now.

Blenux:
-- Replaced the manual duplicated codebase of the strip classes (which can get outdated), and instead just directly import the Strip Properties contents.


Draise14:
Much easier to maintain. Nice.

PR #6782, opened 2026-08-28
3D View - Add - Lattice menu needs icon #6781

frecklesarte:
in line 3550 of DIR: space_view_3d.py :  layout.menu("VIEW3D_MT_lattice_add"), was needing an icon, I added: ", icon="OUTLINER_OB_LATTICE")" in DIR: space_view_3d.py effectively fixing the missing icon.

buld and debug succeeded.

Github-Repo issue title has a typo, change "lattive" to "lattice".

image

Issue #6767, opened 2026-08-19
Linux - Update icon system for an error in DEB installer (conflicts with Blender)


marco62g-ui:
Problem installing the deb package in Debian Trixie due to the inability to overwrite the Blender icon during installation.
I would recommend changing the icon name, at least in the Debian package, to bforartists.svg; it seems the easiest thing to do.

dpkg: error processing archive ~/Desktop/bforartists_5.2.0.deb (--install):
Attempt to overwrite '/usr/share/icons/hicolor/scalable/apps/blender.svg', which is also included in the blender-data package (4.3.2+dfsg-2).

Hi!

Draise14:
Good call.

PR #6789, opened 2026-08-31
Development: Fix broken Git LFS pointers

Quackarooni:
This was causing issues in https://github.com/Bforartists/Bforartists/pull/6782.

These two files got committed as raw binary files instead of Git LFS pointers, causing mismatches when other people pull in from `master`.


Draise14:
Makes sense, I am not sure I have done a merge on linux and these are linux related files, so some dependencies may have not been merged correctly from Windows. 

Being two, it seems like an easy fix.

Draise14:
I wonder if this solves this: #6767

Quackarooni:
> I wonder if this solves this: #6767

I don't know for sure, but the file names are different, so I lean towards not quite.

Issue #6780, opened 2026-08-26
3D Sequencer - Fix timeline gizmos to retime shots


Draise14:
<img width="1757" height="475" alt="Image" src="https://github.com/user-attachments/assets/695a143a-6a02-44e2-acda-31bf3d3519fd" />

We used to be able to adjust strip timing and position from directly in the dopesheet with the gizmo. Now the gizmo is only decorative. 

# Tasks
1. Fix usability of the gizmo
2. Make sure start and end positions and sliding work intuitively
3. Fix "margin" so it sits snug on the Marker lines (if they exist)

Related:
 #6779

Draise14:
Ah, The Node Minimap commit (#6466, May 2026) removed the else default branch in WM_gizmomap_drawstep_from_gizmo_group. Every Python gizmo group (like the 3D Sequencer's dopesheet gizmos) now gets an uninitialized drawstep — they're never drawn and never hit-tested. That's why the handles are "decorative only".

Draise14:
Kinda went through and built a wishlist for the sequencer gizmo, better now than never no? This is a massive upgrade

# Scene Strip Gizmo Features:
- Set strip frame ranges, fallback to active scene time range
- If not scene strips, set current scene range
- Can move or slip scene range start and end
- Set preview range
- Set scene and preview range
- Set scene frame range
- Extend or trim ranges
- Show other scene strips and colors, transparency customizable
- Show scene and strip names, customizable
- Bidirectional playhead sync when sync is on

Draise14:
Wondering if the overlay should show only in the story board and 3d sequencer workspace. 

It can be used in normal animation layouts but only if the set frame range or set preview is toggled. So might come in the way.

Draise14:
Will open a new task later. Will merge this after the merge and some client work.

Draise14:
Needs a remerge.

PR #6795, opened 2026-09-04
3D Sequencer: Cleanup

zNightlord:
Cleanup opentimeline io etc..., use relative import, fix gizmos in #6780 

https://github.com/user-attachments/assets/cd4b3741-5fe7-4c48-998e-cc062408038a


Draise14:
Oof. Was working this and rebuilding the gizmos in C. 

Will have to review if the work is double.

zNightlord:
> Oof. Was working this and rebuilding the gizmos in C.
> 
> Will have to review if the work is double.

the gizmo code before function based on flag is if there is there is no "3D" then it's 2D due to that it breaks other Python made gizmos too 
so either having the UI and TOOL flags exposed for Python (complex) or this is one simpler way as revert it to as it is before, there is no "3D" take one of the UI or TOOL 2D.

https://github.com/user-attachments/assets/9080b799-ddb0-4a6e-89a3-873fcf09ac3a

zNightlord:
> > Oof. Was working this and rebuilding the gizmos in C.
> > Will have to review if the work is double.
> 
> the gizmo code before function based on flag is if there is there is no "3D" then it's 2D due to that it breaks other Python made gizmos too so either having the UI and TOOL flags exposed for Python (complex) or this is one simpler way as revert it to as it is before, there is no "3D" take one of the UI or TOOL 2D.
> 

A test between UI and TOOL fallback on other gizmos
image

Draise14:
Merged it into the other work. Will test.

PR #6801, opened 2026-09-06
Node Editor - Add & Swap Menu - Fix duplicated entries in Group menu #6800

### LINKED ISSUE #6800: Node Editor - Add & Swap Menu - Duplicated entries in `Group` menu
Image

Both our `draw_groups` function *(the one with icons)*, and Blender's implementation *(without icons)*, show up in the `Group` submenu. This was likely not caught during merging, but this should be deduplicated.

Quackarooni:
| Before | After |
| --- | --- |
| image | image |

Resolves: #6800

### DIFF
--- a/scripts/startup/bl_ui/node_add_menu.py
+++ b/scripts/startup/bl_ui/node_add_menu.py
@@ -50,7 +50,7 @@ def icon_sorting_function(group):
 
 
 # BFA - Separated function from `draw_node_group_add_menu` so that both the add menu and toolshelf can use this
-def draw_node_groups(context, layout, operator_id="node.add_node", use_transform=True):
+def draw_node_groups(context, layout, operator_id="node.add_node", use_transform=True, is_linked=False):
     space_node = context.space_data
     node_tree = space_node.edit_tree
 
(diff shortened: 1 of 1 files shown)


Draise14:
I think we had these at a top level for a reason, things were getting kinda deep for usability reasons, and to the sidebar they also are top level. So consistency here with usability and with less deep accessibility kept in mind will help. 

The "Linked" sub-menu can be a label if necessary, but those entries should go higher one level like from the sidebar. 

If we do put them under a sub-label or menu, then the sidebar should also have a sub-menu or label

image

Quackarooni:
@Draise14 - I'll just revert it to be back to the old one. 
The `Linked` menu was a recent addition, and I thought it was wanted.
But just keeping it like how we used to makes the code simpler.

We don't have as big of a need to distinguish them with a separate menu like in Blender,
since the icons and grouping are good enough differentiators already.

Draise14:
Nice. Now I see why the initial! I wrote "wip" after a merge but it was to see why this happened and report a fix for it. But the fix was removal. Sorry bout that. 

Thanks! You came back with force. 🤩

Quackarooni:
@Draise14 
> I wrote "wip" after a merge but it was to see why this happened and report a fix for it. But the fix was removal. Sorry bout that.

No worries, that's understandable! ^^
And yeah, hahah been kinda needing a break from the Clay Pencil stuff as of late, so figured I could grind out a couple of BFA tasks just for a change of pace. 😄

PR #6812, opened 2026-09-14
Properties Editor - Render - EEVEE - Performance - Double column issue #6754

### LINKED ISSUE #6754: Properties Editor - Render - EEVEE - Performance - Double column issue
Image

In the High Quality Normals and Anisotropy, they are on the same row, and should be on two lines. Not two columns.

frecklesarte:
Added 2 Columns, aligned decorators, cleaned up codebase.

Image

image

Anisotropic filter is now on its own column, cleaned up the repeat decorator code, added a column separator in-between the 2 columns to match the UI style of other property sections. turned on row.use_property_split, so "anisotropic filter" is not easily hidden when the properties panel is resized.


Draise14:
On fire!!!! ❤️‍🔥

Quackarooni:
@Draise14 My bad for not mentioning it, but I was gonna review the code this morning. 😅
IMO it could've been substantially simpler. 

Also I take some issue with the commit & PR title not being cleaned up, it's a bit hard to read through in the git history.

Draise14:
Yeah guess the extra splits declars are repeated unnecessarily. 

Guess we can reopen and repull.

But alas reminds me of my first commits. No biggie!

scripts/startup/bl_ui/properties_render.py
```diff
@@ -872,14 +872,19 @@ def draw(self, context):
...
+        col = layout.column(align=True) # BFA
+        
+        row = col.row()
+        row.use_property_split = True
         row.prop(rd, "use_high_quality_normals")
-        row.prop_decorator(rd, "use_high_quality_normals")
+        col.separator( factor = 1)
+        
+        row = col.row()
+        row.use_property_split = True
```
Quackarooni, line 886: `use_property_split` and `use_property_decorate` can be inherited from parent layouts.
So this wouldn't be necessary to define again unless we want this specific row to be different.

scripts/startup/bl_ui/properties_render.py
```diff
@@ -872,14 +872,19 @@ def draw(self, context):
...
 
-        layout.use_property_split = False
-        layout.use_property_decorate = False  # No animation.
+        layout.use_property_split = True
+        layout.use_property_decorate = True  # No animation.
 
-        row = layout.row() # BFA
+        col = layout.column(align=True) # BFA
+        
+        row = col.row()
```
Quackarooni, line 880: `col.row()` isn't mandatory here, calling `col.prop()` would've added each prop in a new row too.
`.row()` is only really necessary if you're planning to put more than one item in a single row.

scripts/startup/bl_ui/properties_render.py
```diff
@@ -872,14 +872,19 @@ def draw(self, context):
...
+        row.use_property_split = True
         row.prop(rd, "use_high_quality_normals")
-        row.prop_decorator(rd, "use_high_quality_normals")
+        col.separator( factor = 1)
```
Quackarooni, line 883: Using `separator` to space out items in a layout where `align=True` is not ideal.
Usually, try `align=False` first.

Or for panels, directly use `layout.prop` instead of `col.prop`, for some reason the base layout has a tad bit more spacing than column layouts.

Quackarooni:
@Draise14 I can prolly just open a new PR for cleanup and point them there.
I left some review notes for reference, but I don't think it'd be great to have them redo it. 😅

PR #6814, opened 2026-09-15
Cleanup: Properties Editor - Render - EEVEE - Performance panel

Quackarooni:
Issues with this layout regarding ##6754 were fixed, but the implementation could've been simplified further.


Quackarooni:
@frecklesarte Tagging you so you can use this as reference.

But yeah, here's the layout this code produces:
image

I also left more granular comments in the original PR itself:
https://github.com/Bforartists/Bforartists/pull/6812/changes
That should show the thought process of trimming down this layout code.

